PR #7, opened 2021-06-15
feat: Add initial arroyo implementation

lynnagara:
This is copied from the snuba/arroyo package


arroyo/retries.py
```diff
@@ -0,0 +1,101 @@
+from abc import abstractmethod
```
fpacifici, line 1: utils

fpacifici (approved):
I think all those util modules should be in a utils package like in snuba. 
Other than that this is ok, as it is copy paste of the entire library from the Snuba code base. I don't see any change.

PR #8, opened 2021-06-15
build: Run pytest in CI

lynnagara:


evanh:
Why tox? I was using that initially for the SDK but GHA can already run the tests against multiple python versions.

lynnagara:
Yeah, the action isn't using tox - I was just using that to run tests locally. I can remove the file it if you think we shouldn't be doing that at all.

evanh:
I think it's beneficial to have local tests work as closely as possible to CI. That's why I removed tox, since then my makefile was running the same command as GHA. It also simplifies the repo a bit. However if running tests against multiple Python versions locally is super important, then I would leave tox.

evanh (approved):

PR #10, opened 2021-06-16
build: Add release workflow and changelog

lynnagara:
The release workflow should trigger a release, publish to pypi and update the changelog.

Warning: This PR is untested; putting this up for feedback on the approach.


.craft.yml
```diff
@@ -0,0 +1,13 @@
+minVersion: 0.23.1
+github:
+  owner: getsentry
+  repo: arroyo
+changelog: CHANGES.rst
+changelogPolicy: auto
+statusProvider:
+  name: github
```
fpacifici, line 8: I think we may need to add a context for relying on github status.
https://github.com/getsentry/cdc/pull/38
and the fix for that https://github.com/getsentry/cdc/pull/39

fpacifici (commented):
One comment on the craft file, otherwise it looks fine to me

fpacifici:
Do you also plan to add an integration test that uses Kafka for real ?

.craft.yml
```diff
@@ -0,0 +1,13 @@
+minVersion: 0.23.1
+github:
+  owner: getsentry
+  repo: arroyo
```
BYK, line 4:
```suggestion
minVersion: 0.24.2
```

.craft.yml
```diff
@@ -0,0 +1,13 @@
...
+changelog: CHANGES.rst
+changelogPolicy: auto
+statusProvider:
+  name: github
+artifactProvider:
+  name: github
```
BYK, line 10: No longer needed as this is the default.
```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    name: "Release a new version"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GH_RELEASE_PAT }}
+      - name: Prepare release
+        uses: getsentry/action-prepare-release@v1.1
```
BYK, line 22:
```suggestion
        uses: getsentry/action-prepare-release@v1
```

BYK (approved):
Don't forget the following:

1. Ensure your `main` branch protection does **NOT** have "Include admins" box checked
2. Ensure @getsentry/release-bot is added as an admin to this repo
3. Ensure either `@getsentry/engineering` has write access to this repo or @getsentry-release has write access

.craft.yml
lynnagara: Removed statusProvider; I think we can just rely the defaults here like our SDKs do

lynnagara:
> Do you also plan to add an integration test that uses Kafka for real ?

What kind of test do you have in mind. Btw the `test_kafka.py` tests already rely on Kafka, and will fail if Kafka is not available.

fpacifici:
Ignore that, I did not see we were running Kafka in ci 
https://github.com/getsentry/arroyo/blob/main/.github/workflows/ci.yml#L66-L75

fpacifici (approved):

PR #20, opened 2021-06-25
build: Provide Python wheel and source distribution for easier installation

lynnagara:


fpacifici (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
+name: build
+
+on:
+  push:
+    branches:
+      - release/**
```
BYK, line 6: Why not also on `main` so we get "nightly" builds out of the box?

BYK (approved):
LGTM

PR #21, opened 2021-06-25
build: Rename package to sentry-arroyo

lynnagara:
`arroyo` seems to be unavailable on PyPI. Let's try renaming to
`sentry-arroyo`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ def get_requirements() -> Sequence[str]:
 
 
 setup(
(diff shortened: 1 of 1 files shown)


fpacifici (approved):
meeeh

BYK (approved):
I mean, sure, but we could also say `gully`, or `brook` or something: https://thesaurus.yourdictionary.com/arroyo

🤣

PR #23, opened 2021-06-28
feat: Parallel transform step accepts optional initializer function

lynnagara:
This adds an optional `initializer` argument to the parallel transform step
which allows a custom initialization function to be run when a process is
spawned. The custom initialize function can also be provided via the
ConsumerStrategyFactory.

The primary motivation for this is to allow Snuba to pass a function that
sets up the Sentry SDK immediately when a process is spawned so that
any logs or errors can be reported to Sentry.

This should not be included in the `0.0.1` release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add optional initializer function to parallel transform step. Supports passing a custom function to be run on multiprocessing pool initialization.
+
 ## 0.0.1
 
 - First release 🎉
-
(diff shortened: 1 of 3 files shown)


fpacifici (commented):
So did you reproduce the fact that the subprocesses were not capable to send anything to sentry?
Also do they log locally ?

lynnagara:
Yes, I didn't find any issues with the actual logging or exception handling in subprocesses. However the spawned process doesn't know anything about Sentry SDK. I was able to reproduce this failure to send to Sentry locally.

fpacifici (commented):
What about the crash of a subprocess for an exception thrown by the processing logic? Does that exception get propagated to the main process with the stack trace and successfully sent to Sentry by the main process?

lynnagara:
Yes, if a subprocess crashes it will already crash the main process. Sentry will attempt to send the error before the process shuts down. I tested this locally and the errors were sent successfully. So this PR doesn't attempt to tackle this at all. This is only to handle non-fatal logging to Sentry from within the subprocess.

fpacifici (approved):

PR #25, opened 2021-07-01
perf(metrics) Add some metrics to the batching consumer

fpacifici:
We do not have a metric on the number of items in the batch during flush and on the time between two flushes.

### DIFF
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -113,7 +113,7 @@ class BatchProcessingStrategy(ProcessingStrategy[TPayload]):
 
         self.__batch: Optional[Batch[TResult]] = None
         self.__closed = False
-        self.__flush_done: Optional[float] = None
+        self.__flush_done = time.time()
 
     def poll(self) -> None:
         """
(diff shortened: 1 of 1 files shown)


arroyo/processing/strategies/batching.py
```diff
@@ -113,6 +113,7 @@ def __init__(
 
         self.__batch: Optional[Batch[TResult]] = None
         self.__closed = False
+        self.__flush_done: Optional[float] = None
```
lynnagara, line 116: Should this also be time.time()?

fpacifici: not really. We do not necessarily start submitting messages right after instantiation.

lynnagara: Aren't we just skipping the recording of the first batch then?

For future batches it looks like we are setting this value on the prior call to `strategy.poll()`, which seems to happens pretty much after instantiation, even if no messages were submitted.

fpacifici: well, this is a good point

lynnagara (approved):

PR #26, opened 2021-07-19
feat: Drop events that fail processing

lynnagara:
If drop_failed_messages is passed to the StreamProcessor it immediately drops
a message that fails processing once.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -110,7 +110,7 @@ class StreamProcessor(Generic[TPayload]):
                     self._run_once()
                 except Exception:
                     # Log 1 in 10 bad events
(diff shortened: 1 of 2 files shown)


vladkluev (approved):
LFIG

nikhars (approved):

arroyo/processing/processor.py
```diff
@@ -110,7 +110,7 @@ def run(self) -> None:
                     self._run_once()
                 except Exception:
                     # Log 1 in 10 bad events
-                    if random.random() > 0.1:
+                    if random.random() < 0.1:
```
vladkluev, line 113: nice

PR #28, opened 2021-09-10
feat: Keep track of timestamps together with offsets in a batch

lynnagara:
This is one of the least invasive ways I've found so far to make the
timestamp associated with an offset availble to the collector processing step.
The end goal of this is to allow Snuba to produce that timestamp to the commit log
during the commit offsets phase, so that it can later be used for subscription scheduling.

We already keep track of the hi and lo offset for each batch and use that to commit offsets.
With this change we will also keep track of the associated timestamp together with it's
corresponding offset number.

### DIFF
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -563,8 +563,6 @@ class KafkaConsumer(Consumer[KafkaPayload]):
 
         assert result is not None  # synchronous commit should return result immediately
 
-        self.__staged_offsets.clear()
-
         offsets: MutableMapping[Partition, Offset] = {}
 
         for value in result:
@@ -582,6 +580,8 @@ class KafkaConsumer(Consumer[KafkaPayload]):
             offsets[partition] = Offset(
                 value.offset, self.__staged_offsets[partition].timestamp
             )
+
+        self.__staged_offsets.clear()
 
         return offsets
 
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -526,21 +526,21 @@ class KafkaConsumer(Consumer[KafkaPayload]):
 
         return [*self.__paused]
 
-    def stage_offsets(self, offsets: Mapping[Partition, Position]) -> None:
+    def stage_positions(self, positions: Mapping[Partition, Position]) -> None:
         if self.__state in {KafkaConsumerState.CLOSED, KafkaConsumerState.ERROR}:
             raise InvalidState(self.__state)
 
-        if offsets.keys() - self.__offsets.keys():
+        if positions.keys() - self.__offsets.keys():
             raise ConsumerError("cannot stage offsets for unassigned partitions")
 
         self.__validate_offsets(
-            {partition: position.offset for (partition, position) in offsets.items()}
+            {partition: position.offset for (partition, position) in positions.items()}
         )
 
         # TODO: Maybe log a warning if these offsets exceed the current
         # offsets, since that's probably a side effect of an incorrect usage
         # pattern?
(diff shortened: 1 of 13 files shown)


arroyo/processing/strategies/batching.py
```diff
@@ -160,10 +160,13 @@ def submit(self, message: Message[TPayload]) -> None:
         self.__metrics.timing("process_message", duration)
 
         if message.partition in self.__batch.offsets:
-            self.__batch.offsets[message.partition].hi = message.next_offset
+            self.__batch.offsets[message.partition].hi = Offset(
+                message.next_offset, message.timestamp
```
lynnagara, line 164: This is probably not right

arroyo/backends/kafka/consumer.py
```diff
@@ -526,21 +526,23 @@ def paused(self) -> Sequence[Partition]:
...
+    def stage_offsets(self, offsets: Mapping[Partition, Offset]) -> None:
         if self.__state in {KafkaConsumerState.CLOSED, KafkaConsumerState.ERROR}:
             raise InvalidState(self.__state)
 
         if offsets.keys() - self.__offsets.keys():
             raise ConsumerError("cannot stage offsets for unassigned partitions")
 
-        self.__validate_offsets(offsets)
+        self.__validate_offsets(
+            {key: offset.offset for (key, offset) in offsets.items()}
```
nikhars, line 537: Nit: Would be easier to read code if you renamed "key" to "partition" wherever Offset is used.

arroyo/types.py
```diff
@@ -63,3 +63,9 @@ def __repr__(self) -> str:
         # ``__slots__`` for performance reasons. The class variable names
         # would conflict with the instance slot names, causing an error.
         return f"{type(self).__name__}(partition={self.partition!r}, offset={self.offset!r})"
+
+
+@dataclass
+class Offset:
+    offset: int
```
nikhars, line 70: Nit: Reading code like offset.offset is harder to read. If you rename this to kafka_offset, reading code like offset.kafka_offset would be much more explicit

nikhars (approved):
I only have minor comments. Else code looks good to me.

arroyo/types.py
lynnagara: Thanks, updated to kafka_offset. Naming is hard.

tkaemming: I don't agree with this. This was intentionally written to be independent from Kafka so that it other conceptually similar backend implementations ([local backend for testing](https://github.com/getsentry/arroyo/blob/main/arroyo/backends/local/backend.py), Pulsar, Google PubSub, Redis Streams, PostgreSQL logical slots, etc.) could be substituted with minimal effort and ideally used by other teams in different environments. Referencing Kafka by name here clearly breaks that encapsulation.

"Offset" may have been too loaded of a term to begin with in the original method signatures (for the same reason — it is too implicitly coupled to a specific backend implementation.) Maybe one of "Cursor" or "Position" would have been more appropriate in the first place. I think `offset.offset` is more of an indication that the conceptual model has changed so that this no longer represents a simple offset but something different.

Aside, considering this data structure is seemingly immutable, it should probably be a `NamedTuple` for efficiency and immutability (or a dataclass with `__slots__` defined and `frozen=True`.)

lynnagara: Good point, `Position` makes sense to me as a container holding the offset and timestamp

tkaemming: Seems like `commit_offsets`, `state_offsets`, etc. should also be updated for consistency with the conceptual change.

lynnagara: Updated to stage_positions, commit_positions

tkaemming: Thanks

PR #31, opened 2021-09-17
feat: Temporarily support missing `orig_message_ts` header for rollout

lynnagara:


arroyo/synchronized.py
```diff
@@ -23,11 +23,13 @@ class Commit:
     group: str
     partition: Partition
     offset: int
-    orig_message_ts: datetime
+    orig_message_ts: Optional[datetime]
 
 
 class CommitCodec(Codec[KafkaPayload, Commit]):
     def encode(self, value: Commit) -> KafkaPayload:
+        assert value.orig_message_ts is not None
```
evanh, line 31: If we are allowing None for `orig_message_ts` why do you assert this here?

lynnagara: All new messages must be produced with this timestamp. Line 41 will fail if it is None here. We are only temporarily allowing orig_message_ts to be none when decoding as existing messages produced with an old version of arroyo will be missing the header.

evanh (approved):

PR #32, opened 2021-10-20
perf: Record time spent by synchronized consumer `poll()` method

lynnagara:
Recording a metric to help better understand the impact of synchronized consumer
on subscription performance.

### DIFF
--- a/arroyo/synchronized.py
+++ b/arroyo/synchronized.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from dataclasses import dataclass
 from datetime import datetime
 from threading import Event
@@ -10,6 +11,7 @@ from arroyo.errors import ConsumerError, EndOfPartition
 from arroyo.types import Message, Partition, Position, Topic, TPayload
 from arroyo.utils.codecs import Codec
 from arroyo.utils.concurrent import Synchronized, execute
+from arroyo.utils.metrics import get_metrics
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


arroyo/synchronized.py
```diff
@@ -209,6 +213,12 @@ def unsubscribe(self) -> None:
         return self.__consumer.unsubscribe()
 
     def poll(self, timeout: Optional[float] = None) -> Optional[Message[TPayload]]:
+        start = time.time()
+        res = self.__poll(timeout)
+        self.__metrics.timing("synchronized.poll", time.time() - start)
```
nikhars, line 218: I am curious to understand the usefulness of this metric in isolation. Would you be comparing `synchronized.poll` time with something else? Let's say that the metric returns 40ms. How do you use that information?

fpacifici: This metric needs to be sampled. Otherwise we will add more contention in recording the metric. 

I think what is more useful is the total breakdown of where we spend time, more than the individual data point for each poll operation.

Since the time you are recording here will be vanishingly small, why don't you keep a state in the consumer of all the timings you are looking at, sum each time and then flush them (by recording only the sum) when you flush a batch.
You could flush more often but still not for each message.

This would give you a good average over time. but you will loose granularity if you try to find a p99.

In alternative you can just sample this and record the metric only once every X messages. 
We have something set up for sampling 
https://github.com/getsentry/snuba/blob/master/snuba/settings/__init__.py#L51-L54

fpacifici (changes requested):
Removing the approval. See the comment about sampling the metric

PR #36, opened 2021-12-07
ref(streaming-strategy): Add batch_execute option

MeredithAnya:


arroyo/processing/strategies/streaming/transform.py
```diff
@@ -230,44 +230,83 @@ def parallel_transform_worker_apply(
...
+        chunk = [input_batch[ind] for ind in range(len(input_batch) - start_index)]
         try:
-            result = function(message)
+            result = function(chunk)
         except Exception:
-            # The remote traceback thrown when retrieving the result from the
-            # pool elides a lot of useful data (and usually includes a
-            # truncated traceback), logging it here allows us to get this
-            # information at the expense of sending duplicate events to Sentry
-            # (one from the child and one from the parent.)
```
fpacifici, line 247: Why was this message removed ?

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -230,44 +230,83 @@ def parallel_transform_worker_apply(
...
-            output_batch.append(
-                Message(message.partition, message.offset, result, message.timestamp)
-            )
-        except ValueTooLarge:
-            # If the output batch cannot accept the transformed message when
-            # the batch is empty, we'll never be able to write it and should
-            # error instead of retrying. Otherwise, we need to return the
-            # values we've already accumulated and continue processing later.
-            if len(output_batch) == 0:
+        for ind, message in enumerate(chunk):
```
fpacifici, line 251: This does not seem a clean way to accomplish your goal. 
By introducing a batching concept in the `parallel_transform_worker_apply` function we are making it twice as complex.
The same result can be achieved without touching the infrastructure at all by just passing batches of messages to the parallel transform step:
The message type for the parallel transform step is generic. You can pass whatever you want. 
What about batching messages together in a transform step before the parallel step, passing these batches to the parallel step, and eventually if you need, un-batch them in a step after the parallel step? 

This approach seems to have a few advantages:
- You maximize the gain from the parallel step. Big messages that take long to process parallelize better than smaller messages as the overhead of the parallelization counts less. So batching messages before sending them through the parallel step should help.
- This seems an ad hoc feature that does not need to be in the infrastructure as it is likely to be specific to the product feature you are building. We want to keep the infrastructure abstractions as simple as possible (which are already insanely too complex) so that they are easier to understand. In order for this to type check properly, the `function` passed to `parallel_transform_worker_apply` must be able to support two input types, `TPayload` and `Sequence[TPayload]`, which will have to be supported by all use cases that rely on arroyo.
- This implementation becomes more confusing as now it deals with two independent concepts of batch that mean very different things.

fpacifici (changes requested):
I don't think you should need to touch arroyo at all to accomplish your goal. 
See comment inline about where to do the batching.

MeredithAnya:
Per @fpacifici's comments, this is not the best approach, going to implement a different strategy to use in https://github.com/getsentry/sentry/pull/30462 instead

PR #41, opened 2022-03-02
perf(collect): Add ParallelCollect step

nikhars:
Added a parallel collect step which runs the closing of the batch
in another thread. This would allow new batches to be created and
starting to fill in while the previous batch is still getting closed.

### DIFF
--- a/arroyo/processing/strategies/streaming/collect.py
+++ b/arroyo/processing/strategies/streaming/collect.py
@@ -1,13 +1,24 @@
 import logging
 import time
+from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Callable, Generic, Mapping, MutableMapping, Optional
+from typing import (
+    Awaitable,
+    Callable,
+    Generic,
+    Mapping,
+    MutableMapping,
+    Optional,
+    TypeVar,
+)
 
 from arroyo.processing.strategies.abstract import ProcessingStrategy as ProcessingStep
 from arroyo.types import Message, Partition, Position, TPayload
 
 logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


arroyo/processing/strategies/streaming/collect.py
```diff
@@ -97,54 +98,105 @@ def __init__(
...
+        if self.batch is not None:
+            self.batch.join(timeout)
+            logger.info("Completed processing %r.", self.batch)
+            self.batch = None
+
+
+class ParallelCollectStep(CollectStep[TPayload]):
+    """
+    ParallelCollectStep is similar to CollectStep except it allows the closing and reset of the
+    batch to happen in a threadpool. What is allows is for the next batch to start getting
```
onewland, line 157: I think this should read `What this allows for`

onewland (commented):
Maybe a dumb question, but do we need to be concerned about some kind of special behavior around error handling? I don't see that in the tests. What happens/should happen if commit_function raises an error when called?

arroyo/processing/strategies/streaming/collect.py
```diff
@@ -97,54 +98,105 @@ def __init__(
...
+        self.future: Optional[Future[None]] = None
+
+    def close_and_reset_batch(self) -> None:
+        """
+        Closes the current batch in an asynchronous manner. Waits for previous work to be completed before proceeding
+        the next one. We can provide the existing batch to the threadpool since the collector is going to make a
+        new batch.
+        """
+        if self.future:
+            self.future.result()
```
MeredithAnya, line 182: should this have a timeout at all? not sure how likely this is to raise a `TimeoutError`. Seems like we should def be done submitting the previous batch before a new batch is filled up, but maybe if the consumer is backlogged this could happen?

nikhars:
> Maybe a dumb question, but do we need to be concerned about some kind of special behavior around error handling? I don't see that in the tests. What happens/should happen if commit_function raises an error when called?

Any errors which might happen would be raised by the future. I have added a test which validates that exception is thrown when commit fails. Hope that helps.

arroyo/processing/strategies/streaming/collect.py
nikhars: Added a timeout of 5 seconds. Most often the future's result would be available immediately. But I have added it if for some reason clickhouse is down, we can abort.

onewland (approved):

arroyo/processing/strategies/streaming/collect.py
```diff
@@ -97,54 +98,106 @@ def __init__(
...
+        logger.info("Completed processing %r.", batch)
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        # We should finish the previous batch before proceeding to the finish the existing one.
+        if self.future is not None:
+            self.future.result(timeout)
+
+        self.__threadpool.shutdown()
+
+        super().join(timeout)
```
lynnagara, line 203: Not a huge deal but I think it would be more correct to pass the remaining time here and not the entire timeout.

PR #42, opened 2022-03-03
Familiarized with Arroyo.

rahul-kumar-saini:
# (Not actually contributing this stuff to the Arroyo repo)
### This PR is just meant as a request for comments on how I've used Arroyo stuff

I made a processing strategy which applies a transformation and produces to a topic, given the transform function and producer to produce to.

In the demo (gif) I’m manually producing some auth credentials to `raw-topic`, the password gets hashed (transformation function) and the credentials are produced to `hashed-topic`.
- Bottom left: running the script
- Bottom right: manually producing a message to the raw topic
- Top right: raw topic
- Top left: hashed topic

simple_producer


rahul-kumar-saini:
Thank you @MeredithAnya for help with wiring the pieces together and debugging 😄

onewland:
I wonder if this could be added to something like an `examples/` directory in this repository. I know that I would have found that useful as somebody new to arroyo.

lynnagara:
+1 for adding the `examples` directory. Though I think this example may be a bit complex in its current form. If you make an even more simplified version (and add inline code comments) I think it could be a good candidate for that.

rahul-kumar-saini:
Closing as there is a real PR now #44

PR #44, opened 2022-03-04
Added an example for Arroyo usage.

rahul-kumar-saini:
- There is limited documentation and (simple) examples for Arroyo's usage
- Created an example for usage while learning to work with Arroyo and documented thoroughly.


arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,95 @@
...
+        # KafkaProducer asynchronously produces a message so it returns a "Future"
+        self.__futures.append(MessageToFuture(message, future))
+
+    def close(self) -> None:
+        self.__closed = True
+
+    def terminate(self) -> None:
+        self.close()
+
+    def join(self, timeout: Optional[float] = None) -> None:
```
lynnagara, line 77: We should probably respect the timeout if one is passed

arroyo/examples/transform_and_produce/script.py
```diff
@@ -0,0 +1,53 @@
...
+HASHED_TOPIC = Topic("hashed-topic")
+
+
+if __name__ == "__main__":
+
+    # Create a Producer, KafkaProducer is a wrapper on confluent_kafka.Producer
+    producer = KafkaProducer(
+        {
+            "bootstrap.servers": "localhost:9092",  # Where the broker is, localhost:9092 when running locally
+            "client.id": socket.gethostname(),
```
lynnagara, line 27: Do we need this? Can we leverage the `build_kafka_configuration()` instead?

arroyo/examples/transform_and_produce/script.py
```diff
@@ -0,0 +1,53 @@
...
+RAW_TOPIC = Topic("raw-topic")
+HASHED_TOPIC = Topic("hashed-topic")
+
+
+if __name__ == "__main__":
+
+    # Create a Producer, KafkaProducer is a wrapper on confluent_kafka.Producer
+    producer = KafkaProducer(
+        {
+            "bootstrap.servers": "localhost:9092",  # Where the broker is, localhost:9092 when running locally
```
lynnagara, line 26: Can you define it as a variable at the top of the file, and maybe allow it to be over-ridden via an environment variable or something?

arroyo/examples/transform_and_produce/script.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    # KafkaConsumer is a wrapper on confluent_kafka.Consumer
+    consumer = KafkaConsumer(
+        # A function to bootstrap required configs to create a consumer
+        build_kafka_consumer_configuration(
+            default_config={
+                "bootstrap.servers": "localhost:9092",
+            },
+            auto_offset_reset="latest",  # "define the behavior of the consumer when there is no committed position" - confluent docs
```
lynnagara, line 38: I'd remove this comment, auto_offset_reset means a slightly different thing in Arroyo than confluent kafka

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,95 @@
...
+    to the original consumer.
+    """
+
+    message: Message[KafkaPayload]
+    future: Future[Message[KafkaPayload]]
+
+
+class ProduceStep(ProcessingStep[KafkaPayload]):
+    """
+    Simple step that produces a given message onto a topic.
```
lynnagara, line 34: Agree it's not super complex but I don't think characterizing this as "simple" is particularly helpful to new users. I think we can try to give a more helpful description here.

```suggestion
    Produces a message onto a topic asynchronously and commits the offset of the original message once producing is complete.
```

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,95 @@
...
+    def close(self) -> None:
+        self.__closed = True
+
+    def terminate(self) -> None:
+        self.close()
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        self._commit()
+
+    def _commit(self) -> None:
```
lynnagara, line 80: nit: This function does a bit more than just committing offsets, please name accordingly

lynnagara (commented):
I would like to avoid increasing the size of the library due to the examples, can you please check that these are being excluded from the package?

arroyo/examples/transform_and_produce/README.md
```diff
@@ -0,0 +1,53 @@
...
+- This message is in the form `{"username": "<username>", "password": "<password>"}`
+- The message is submitted to the `HashPasswordStrategy` which hashes the password string
+- The credentials are then submitted to the `ProduceStep` which simply produces the given message to `hash-topic`
+- The `ProduceStep` is also responsible for commiting the original offset back for the consumer
+
+### Usage
+
+---
+
+(Make sure to have `kcat` installed, and read up on usage [here](https://docs.confluent.io/platform/current/app-development/kafkacat-usage.html))
```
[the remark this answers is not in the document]
rahul-kumar-saini: Well `kcat` seems like an essential tool for working with Kafka, I don't even know how I'd monitor messages without it.

arroyo/examples/transform_and_produce/hash_password_strategy.py
```diff
@@ -0,0 +1,69 @@
...
+        # Expected format of the message is {"username": "<username>", "password": "<password>"}
+        auth = json.loads(message.payload.value)
+        hashed = self._hash_password(auth["password"])
+        data = json.dumps({"username": auth["username"], "password": hashed}).encode(
+            "utf-8"
+        )
+
+        payload = KafkaPayload(
+            message.payload.key, data, headers=message.payload.headers
+        )
```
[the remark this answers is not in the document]
rahul-kumar-saini: No reason, just didn't know what to put there.

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,95 @@
...
+class MessageToFuture(NamedTuple):
+    """
+    Map a submitted message to a Future returned by the Producer.
+
+    This is useful for being able to commit the latest offset back
+    to the original consumer.
+    """
+
+    message: Message[KafkaPayload]
+    future: Future[Message[KafkaPayload]]
```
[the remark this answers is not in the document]
rahul-kumar-saini: The way I understand it `message` is the original message from the consumer, `future` is the Future of the message produced by the producer. These messages could have different offsets so the original message is needed for when we commit.

arroyo/examples/transform_and_produce/produce_step.py
rahul-kumar-saini: How should I use the timeout here? The join method calls the method that commits but I'm not sure how to time that and stop processing after a given timeout.

arroyo/examples/transform_and_produce/produce_step.py
rahul-kumar-saini: It's committing the offsets for futures that are done, do you mean something along the line `_commit_and_prune_futures()`?

arroyo/examples/transform_and_produce/produce_step.py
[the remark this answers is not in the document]
lynnagara: Why would the messages have different offsets though? On line 43 of hash_password_strategy.py it seems like you are copying the offsets and the rest of the metadata from the original message into the one that is in the future so they would be the same.

arroyo/examples/transform_and_produce/README.md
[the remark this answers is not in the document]
lynnagara: Although kcat is more powerful, I don't think you're using any features that you can't do with the built in `kafka-console-consumer` for example.

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Commit the latest offset that has its corresponding produce finished
+        if commitable is not None:
+            self.__commit(
+                {
+                    commitable.partition: Position(
+                        commitable.offset, commitable.timestamp
+                    )
+                }
+            )
```
lynnagara, line 99: How come you're only committing the last future?

arroyo/examples/transform_and_produce/produce_step.py
lynnagara: Well, this method does a while loop that checks each future in the queue is done. You could abort that when the timeout is reached.

arroyo/examples/transform_and_produce/script.py
```diff
@@ -0,0 +1,58 @@
...
+from arroyo.examples.transform_and_produce.factory import (
+    HashPasswordAndProduceStrategyFactory,
+)
+
+
+RAW_TOPIC = Topic("raw-topic")
+HASHED_TOPIC = Topic("hashed-topic")
+
+# Where the broker is, localhost:9092 when running locally
+BOOTSTRAP_SERVERS = os.environ.get("BOOTSTRAP_SERVERS") or ["localhost:9092"]
```
lynnagara, line 24: How does this work if set via environment variable? Does bootstrap servers have to be an array?

lynnagara:
> I would like to avoid increasing the size of the library due to the examples, can you please check that these are being excluded from the package?

Any thoughts on this @rahul-kumar-saini?

arroyo/examples/transform_and_produce/produce_step.py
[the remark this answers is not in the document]
rahul-kumar-saini: When we take that message and submit it to `ProduceStep`, we only pass a payload to the producer (L66-L68):

https://github.com/getsentry/arroyo/blob/618b869b732dcb7ee834242d5ee79cff927382c3/arroyo/examples/transform_and_produce/produce_step.py#L63-L70

So what the producer returns is a `Future` for the _new_ message it produces onto the destination topic no? If so, that new message could have a different offset than the original message that was submitted to `ProduceStep`.

arroyo/examples/transform_and_produce/script.py
rahul-kumar-saini: Technically `bootstrap_servers` has the type `Sequence[str]` which tells me it should be a list, tuple etc.
https://github.com/getsentry/arroyo/blob/3d83a4bed1b82e8a1d3b2c376fa23c1721f63606/arroyo/backends/kafka/configuration.py#L43
It just works if I pass a  `str` anyway, so a basic string as an environment variable should work. I put the default in a list to respect the type but I can change it if that doesn't make sense.

rahul-kumar-saini:
> > I would like to avoid increasing the size of the library due to the examples, can you please check that these are being excluded from the package?
> 
> Any thoughts on this @rahul-kumar-saini?

I've updated `setup.py` to exclude the examples module so it should not be part of the package now.
https://github.com/getsentry/arroyo/blob/be0d5672a15961d7b8938d9903c63187e2cbfd22/setup.py#L21

I attempted to bring the `/examples` folder out to the root directory but I'm getting a strange error: `ModuleNotFoundError: No module named 'examples.transform_and_produce'`

I've created `__init__.py` files in all directories that need it, and I believe the modules are recognized at compile time but not at runtime for some reason.

arroyo/examples/transform_and_produce/script.py
lynnagara: If it's supposed to be a list, why not just make it a list? Even though it might just happen to work, I don't think we should rely on that.

lynnagara:
> > > I would like to avoid increasing the size of the library due to the examples, can you please check that these are being excluded from the package?
> > 
> > 
> > Any thoughts on this @rahul-kumar-saini?
> 
> I've updated `setup.py` to exclude the examples module so it should not be part of the package now.
> 
> https://github.com/getsentry/arroyo/blob/be0d5672a15961d7b8938d9903c63187e2cbfd22/setup.py#L21
> 
> I attempted to bring the `/examples` folder out to the root directory but I'm getting a strange error: `ModuleNotFoundError: No module named 'examples.transform_and_produce'`
> 
> I've created `__init__.py` files in all directories that need it, and I believe the modules are recognized at compile time but not at runtime for some reason.

Thanks, this seems fine where it is now.

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,108 @@
...
+        commitable: MutableMapping[Partition, Message[KafkaPayload]] = {}
+
+        while self.__futures and self.__futures[0].future.done():
+            message, _ = self.__futures.popleft()
+            # overwrite any existing message as we assume the deque is in order
+            # committing offset x means all offsets up to and including x are processed
+            commitable[message.partition] = message
+
+            if timeout is not None and time.perf_counter() - start > timeout:
+                return
```
lynnagara, line 99: I think you should commit the last offsets before returning here.

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,108 @@
...
+
+        # Commit the latest offset that has its corresponding produce finished, per partition
+        if commitable is not None:
```
lynnagara, line 102: Under what conditions would it ever be None? It looks like it is always a dictionary.

arroyo/examples/transform_and_produce/produce_step.py
```diff
@@ -0,0 +1,108 @@
...
+        self._commit_and_prune_futures(timeout)
+
+    def _commit_and_prune_futures(self, timeout: Optional[float] = None) -> None:
+        """
+        Commit the latest offset of any completed message from the original
+        consumer.
+        """
+        start = time.perf_counter()
+
+        commitable: MutableMapping[Partition, Message[KafkaPayload]] = {}
```
lynnagara, line 90: nit: I think it should be spelled with double t. ie. `committable`

arroyo/examples/transform_and_produce/produce_step.py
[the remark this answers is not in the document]
lynnagara: Ah sorry, didn't realise it's the message from the producer.

lynnagara (approved):
Looks good to me. Thanks for kicking off `examples` 🎉

PR #46, opened 2022-03-09
ref(metrics): Add metrics for time spent polling and closing batch

nikhars:
Added metrics to account for time spent polling and closing and resetting of batch to observe the current behavior and possible improvements with parallel collect.

### DIFF
--- a/arroyo/processing/strategies/streaming/collect.py
+++ b/arroyo/processing/strategies/streaming/collect.py
@@ -7,6 +7,7 @@ from typing import Callable, Generic, Mapping, MutableMapping, Optional
 
 from arroyo.processing.strategies.abstract import ProcessingStrategy as ProcessingStep
 from arroyo.types import Message, Partition, Position, TPayload
+from arroyo.utils.metrics import get_metrics
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


arroyo/processing/strategies/streaming/collect.py
```diff
@@ -184,6 +200,8 @@ def close_and_reset_batch(self) -> None:
 
         self.future = self.__threadpool.submit(self.__finish_batch, batch=self.batch)
         self.batch = None
+        self._metrics.timing("collect.poll.time", self._collect_poll_time)
+        self._collect_poll_time = 0
```
fpacifici, line 204: Not following why you are recording hte metric only here. 
Why not just collect it in the poll method every like 1-2 seconds ?
Like Meredith did here 
https://github.com/getsentry/sentry/blob/master/src%2Fsentry%2Fsentry_metrics%2Fmultiprocess.py#L285-L295

nikhars: The reason I did it this way is to capture the breakdown of polling with respect to a batch. The current method would allow me to see how much percentage of time we spend doing `polling` and `closing batch` with respect to batch time. Let me take metrics as an example. If the batch time is 16 seconds. And polling takes 0.5s and closing takes 0.5s I can get the percentage easily.

If I did something similar to Meredith's implementation, with reporting of metric every 5 seconds and the batch closing time is 16 seconds, there would be 3 samples where the `polling` time with `closing batch` operation won't be there.

I don't have a strong opinion on this. If you feel we should report every fixed interval I can modify it that way. Let me know which way you think is better.

fpacifici (approved):
ok

PR #47, opened 2022-03-09
Dead Letter Queue

rahul-kumar-saini:
### Overview
- Introduced a module for a [Dead Letter Queue](https://www.enterpriseintegrationpatterns.com/patterns/messaging/DeadLetterChannel.html)
- Added an interface for a DLQ Policy which decides what to do with dead/invalid messages
- Implemented a few sample policies to ignore, raise, or raise after a certain number


arroyo/dead_letter_queue/dead_letter_queue.py
```diff
@@ -0,0 +1,54 @@
...
+    def close(self) -> None:
+        self.__closed = True
+
+    def terminate(self) -> None:
+        self.close()
+        logger.debug("Terminating %r...", self.__next_step)
+        self.__next_step.terminate()
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        self.__next_step.close()
```
evanh, line 53: Why do you close the next step here? Wouldn't the next step close itself in the `join`?

arroyo/dead_letter_queue/policies/count.py
```diff
@@ -0,0 +1,18 @@
...
+class CountInvalidMessagePolicy(DeadLetterQueuePolicy[TPayload]):
+    def __init__(self, limit: int) -> None:
+        self.__limit = limit
+        self.__count = 0
+
+    def handle_invalid_message(
+        self, message: Message[TPayload], e: InvalidMessage
+    ) -> None:
+        self.__count += 1
+        if self.__count > self.__limit:
```
evanh, line 17: Does the count reset if the exception gets hit?

arroyo/dead_letter_queue/dead_letter_queue.py
rahul-kumar-saini: In the description of the join method in the abstract class it says `This method should be called after ``close`` to provide a graceful shutdown`, and I've seen a few examples of `self.__next_step.close()` in `join` methods of other implementations of ProcessingStrategy. I haven't seen any join method call its own close method though.

arroyo/dead_letter_queue/policies/count.py
rahul-kumar-saini: For now, this policy just allows a certain number of `InvalidMessage`'s till it starts raising all incoming exceptions. If we reset the count, this policy changes to be more like `raise every nth exception` instead of `raise after n exceptions`. The only way to reset would be by restarting this DLQ? I'm not sure how to tell it to reset the count without potentially a runtime config or just changing the policy to resetting after an exception is hit.

tests/dead_letter_queue/test_dead_letter_queue.py
```diff
@@ -0,0 +1,69 @@
...
+            raise InvalidMessage
+
+
+def test_dlq() -> None:
+
+    processing_step = FakeProcessingStep()
+
+    dlq_raise = DeadLetterQueue(processing_step, RaiseInvalidMessagePolicy())  # type: ignore
+    dlq_ignore = DeadLetterQueue(processing_step, IgnoreInvalidMessagePolicy())  # type: ignore
+    dlq_count = DeadLetterQueue(processing_step, CountInvalidMessagePolicy(5))  # type: ignore
```
rahul-kumar-saini, line 47: Not sure why, mypy kept complaining: `Cannot infer type argument 1 of "DeadLetterQueue"`.

rahul-kumar-saini, line 47: I'm not sure why mypy kept complaining on this: `Cannot infer type argument 1 of "DeadLetterQueue"`.

rahul-kumar-saini:
I'm unsure how to add metrics to the policies, for now I added a `logger.warning()` on bad messages at the DLQ level but I feel every usage of the DLQ would want its own set of metrics. I think a DLQ could be initialized with a 'name' or something which could then be used in Arroyo for Datadog metrics, unless anyone has a suggestion for a better way.

arroyo/dead_letter_queue/policies/count.py
rahul-kumar-saini: Revamped the entire policy to work as a limit per time unit.

arroyo/dead_letter_queue/dead_letter_queue.py
evanh: Oh I see. OK that makes sense.

evanh (commented):
LGTM. I'll let someone more familiar with arroyo approve though.

arroyo/dead_letter_queue/dead_letter_queue.py
```diff
@@ -0,0 +1,54 @@
...
+    def submit(self, message: Message[TPayload]) -> None:
+        assert not self.__closed
+        try:
+            self.__next_step.submit(message)
+        except InvalidMessage as e:
+            logger.warning(
+                f"Invalid Message caught by Dead Letter Queue: {message}",
+                exc_info=True,
+            )
+            self.__policy.handle_invalid_message(message, e)
```
rahul-kumar-saini, line 42: When we want to start producing the invalid message to a dead letter topic, it could happen here. If the handler call doesn't raise an exception, produce the message to the dead letter topic.

Is there a case where we do not raise an exception and also do not want the message to be forwarded? If so, the handler could return a boolean telling the main class to forward the message or not.

lynnagara (commented):
Just a few comments on the count policy

arroyo/dead_letter_queue/policies/count.py
```diff
@@ -0,0 +1,70 @@
...
+    A saved state in the form `[(<timestamp: int>, <hits: int>), ...]` can be passed
+    on init to load the previously saved state. This state should be aggregated to
+    1 second buckets.
+    """
+
+    def __init__(
+        self,
+        limit: int,
+        seconds: int = 60,
+        load_state: Sequence[Tuple[int, int]] = [],
```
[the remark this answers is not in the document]
rahul-kumar-saini: Same reason as below, if the consumer restarts we can load the state of hits from Redis into this policy.

arroyo/dead_letter_queue/factory.py
```diff
@@ -0,0 +1,30 @@
...
+    ProcessingStrategyFactory,
+)
+from arroyo.types import Partition, Position, TPayload
+
+
+class DeadLetterQueueFactory(ProcessingStrategyFactory[TPayload]):
+    """
+    DLQ Factory which wraps a given ProcessingStrategyFactory with
+    a given DLQ Policy.
+    """
```
rahul-kumar-saini, line 16: From what I've seen so far, we don't directly use processing strategies in Snuba. We do something like:
```python
processing_strategy_factory = SomeProcessingStrategyFactory()
processor = StreamProcessor(
    consumer,
    topic,
    processing_strategy_factory,
)

processor.run()
```

I think since the DLQ is defined as a wrapper on a processing strat/step, we can't really directly use it but instead have to pass some sort of factory to the stream processor. With this DLQ factory we can do something like this instead:

```python
processing_strategy_factory = SomeProcessingStrategyFactory()
processor = StreamProcessor(
    consumer,
    topic,
    DeadLetterQueueFactory(processing_strategy_factory, SomeDLQPolicy()),
)

processor.run()
```

lynnagara: Isn't it already easy enough to incorporate the DeadLetterQueue strategy in an existing factory. i.e. just a matter of adding something like this in your `SomeProcessingStrategyFactory`
```
DeadLetterQueueFactory(strategy, SomeDLQPolicy())
```

arroyo/dead_letter_queue/policies/count.py
```diff
@@ -0,0 +1,70 @@
...
+        add_hit_callback: Optional[Callable[[int], None]] = None,
```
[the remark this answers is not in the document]
lynnagara: I see. Did you consider just adding the more complex CountBackedByRedis policy in Snuba directly and letting it manage Redis state since it seems policies are designed to be extensible. Just a thought as this seems to me rather customized to Snuba's requirements already. The other thing is that we would need to get a fresh count to pass to this strategy not just when the consumer restarts but also in the case of rebalances. That part may be easier if it's all in the strategy as this can be done in the constructor of the policy class.

arroyo/dead_letter_queue/factory.py
rahul-kumar-saini: We could do that but I'm guessing we don't wanna change the fundamental strategy factories like `BatchProcessingStrategyFactory`, but we instead want to pick and choose which consumers we want a DLQ on?

lynnagara: Is it a requirement for the strategies that use the `BatchProcessingStrategyFactory` to work with DLQ. As far as I can tell those ones never raise the `InvalidMessage` error so would never trigger the policy.

arroyo/dead_letter_queue/policies/count.py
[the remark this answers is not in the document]
rahul-kumar-saini: I did have 2 separate classes, a `CountPolicy` and a `StatefulCountPolicy` (which extended the former and had the callback logic), but I had them both in Arroyo so I merged them together. I didn't consider putting the Redis backed count policy in Snuba but it makes sense, I'll split it back up.

Why do we need to refresh the count on a rebalance? Wouldn't we want to keep track of total number of invalid messages regardless of consumer?

arroyo/dead_letter_queue/factory.py
rahul-kumar-saini: It's not a requirement, just an example of where the DLQ could be used. Nothing raises the `InvalidMessage` exception as of now, but when the DLQ is used we'd change the relevant processor to raise the exception so the DLQ can handle stuff appropriately.

lynnagara: Ok. Because I would alternatively suggest just passing it here instead and doing essentially the same thing we do with the optional filter step, e.g.  https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/streaming/factory.py#L126-L127.

arroyo/dead_letter_queue/policies/count.py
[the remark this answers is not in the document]
lynnagara: It's just that a rebalance causes the strategy to get recreated. I assumed that would recreate the policy so it needs to get a different count passed in `load_state` (the latest one)

tests/dead_letter_queue/test_dead_letter_queue.py
```diff
@@ -0,0 +1,189 @@
...
+    invalid_payload = KafkaPayload(None, b"", [])
+    return Message(Partition(Topic(""), 0), 0, invalid_payload, datetime.now())
+
+
+def test_raise(
+    valid_message: Message[KafkaPayload],
+    invalid_message: Message[KafkaPayload],
+    processing_step: FakeProcessingStep,
+) -> None:
+    dlq_raise = DeadLetterQueue(processing_step, RaiseInvalidMessagePolicy())  # type: ignore
```
lynnagara, line 75: How come `type: ignore` was needed? Can we fix instead?

arroyo/dead_letter_queue/dead_letter_queue.py
```diff
@@ -0,0 +1,54 @@
+import logging
```
fpacifici, line 1: I think this entire module should be under `arroyo/processing/strategies/`. Then you can export the symbols you need like this https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/streaming/__init__.py#L1-L13

arroyo/dead_letter_queue/dead_letter_queue.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def submit(self, message: Message[TPayload]) -> None:
+        assert not self.__closed
+        try:
+            self.__next_step.submit(message)
+        except InvalidMessage as e:
+            logger.warning(
+                f"Invalid Message caught by Dead Letter Queue: {message}",
+                exc_info=True,
+            )
```
fpacifici, line 41: We cannot have a warning every time. Let the policy decide whether to log. 
If the dead letter queue ends up receiving InvalidMessage exceptions for all messages on a high volume topic we would be floodding ourselves with messages to log (they all go in sentry).
So Arroyo does not have enough context to decide if an invalid message should be logged or not. the policy can have such context.
Recording a metric is a better option.

arroyo/dead_letter_queue/policies/abstract.py
```diff
@@ -0,0 +1,23 @@
+from abc import ABC, abstractmethod
+from typing import Generic
+
+from arroyo.types import Message, TPayload
+
+
+class InvalidMessage(Exception):
+    pass
```
fpacifici, line 8: Why not providing the message as an attribute of the exception ?

arroyo/dead_letter_queue/policies/raise_e.py
```diff
@@ -0,0 +1,15 @@
+from arroyo.dead_letter_queue.policies.abstract import (
+    DeadLetterQueuePolicy,
+    InvalidMessage,
+)
+from arroyo.types import Message, TPayload
+
+
+class RaiseInvalidMessagePolicy(DeadLetterQueuePolicy[TPayload]):
+    def __init__(self) -> None:
+        pass
```
fpacifici, line 10: no need for this as it does not initialize anything

fpacifici (changes requested):
Please see the comments inline

arroyo/dead_letter_queue/dead_letter_queue.py
rahul-kumar-saini: It's more intuitive to me to have the DLQ be responsible for receiving and then forwarding the message, and the policy possibly interrupts this process by raising an exception.

This way the DLQ could be initialized with a topic to produce to, it feels off to have the policy initialized with a topic to produce a message to. The "produce to dead letter topic" logic would also have to be repeated across policies (unless a base Policy class implements this)

tests/dead_letter_queue/test_dead_letter_queue.py
rahul-kumar-saini: I keep getting a `Cannot infer type argument 1 of "DeadLetterQueue"` mypy error, I'm not sure why. Is there something up with how I've defined the Fake step/policy classes?

arroyo/processing/strategies/dead_letter_queue/policies/count.py
```diff
@@ -0,0 +1,60 @@
...
+    This window is 1 minute by default. The exception associated with the invalid
+    message is raised for all incoming invalid messages which go past this limit.
+
+    A saved state in the form `[(<timestamp: int>, <hits: int>), ...]` can be passed
+    on init to load the previously saved state. This state should be aggregated to
+    1 second buckets.
+    """
+
+    def __init__(
+        self, limit: int, seconds: int = 60, load_state: Sequence[Tuple[int, int]] = []
```
lynnagara, line 34: Do we still need load_state? I thought we're having a separate policy that does loading from Redis now?

tests/dead_letter_queue/test_dead_letter_queue.py
lynnagara: Why don't you specify the type?
```
dlq_raise: DeadLetterQueue[KafkaPayload] = DeadLetterQueue(processing_step, RaiseInvalidMessagePolicy())
```

arroyo/processing/strategies/dead_letter_queue/policies/count.py
rahul-kumar-saini: I'm thinking that separate policy should inherit this policy, so for it to load a state when initialized I made this one loadable.

So it could do something like:
```python3
def __init__(...):
    super().__init__(limit, seconds, load_state)
    self._init_redis_stuff()
```

Otherwise there would need to be some method in the base class to load a state or something. The other option is to just not inherit from this class at all.

tests/dead_letter_queue/test_dead_letter_queue.py
rahul-kumar-saini: Ah ok thanks 🤦‍♂️

arroyo/processing/strategies/dead_letter_queue/policies/count.py
lynnagara: I see. One option would be to make `self.hits` not private. Or you can just keep it this way, but if you do please make it:
```
load_state: Optional[Sequence[Tuple[int, int]]] = None
```

due to https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
```diff
@@ -0,0 +1,54 @@
...
+        next_step: ProcessingStep[TPayload],
+        policy: DeadLetterQueuePolicy[TPayload],
+    ) -> None:
+        self.__next_step = next_step
+        self.__policy = policy
+        self.__closed = False
+        self.__metrics = get_metrics()
+
+    def poll(self) -> None:
+        self.__next_step.poll()
```
fpacifici, line 34: Thinking over this, we want to intercept the `InvalidMessage` exception here as well. 
Most consumers only do useful stuff per message in submit, but there are some that only accumulate messages in submit and do the work in `poll`

lynnagara (approved):

fpacifici (approved):

PR #48, opened 2022-03-21
fix(dlq): Export InvalidMessage from DLQ

rahul-kumar-saini:
🤮 `from arroyo.processing.strategies.dead_letter_queue.policies.abstract import InvalidMessage`

😎 `from arroyo.processing.strategies.dead_letter_queue import InvalidMessage`


lynnagara (approved):

lynnagara:
I do kind of wonder if this should be within the DLQ module at all or on the top level

PR #49, opened 2022-03-22
fix(dlq): Removed Generic payload from DLQ Policy

rahul-kumar-saini:
### tl;dr
A DLQ Policy does not care about the Payload type of a Message.

### context
Originally a Policy itself would take the message as a parameter and the signature for the handler looked like: `handle_invalid_message(message: Message[TPayload], e: InvalidMessage)`.

That was changed in favor of just having the exception itself hold the message. Removing the generic was forgotten about when the signature changed to `handle_invalid_message(e: InvalidMessage)`.

### DIFF
--- a/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
+++ b/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
@@ -23,7 +23,7 @@ class DeadLetterQueue(ProcessingStep[TPayload]):
     def __init__(
         self,
         next_step: ProcessingStep[TPayload],
-        policy: DeadLetterQueuePolicy[TPayload],
+        policy: DeadLetterQueuePolicy,
     ) -> None:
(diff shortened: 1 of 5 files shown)


lynnagara (approved):
May be worth adding a line to the changelog too

rahul-kumar-saini:
> May be worth adding a line to the changelog too

The bot auto adds a line with the PR name and adds the version info, where/what could I manually add without messing up the bots workflow?

lynnagara:
Ah, just realised one is automatically generated now if it's empty (didn't used to be). So if you want to go with that and not modify the changelog file itself all good. But the way to do it would be to add an entry to the changelog file under the heading `Unreleased` and write your changes under that. Craft will update the version number next time the publish workflow is run. The `auto` changelog policy is described here https://github.com/getsentry/craft/blob/47468020b3576960947b7353cb3eb6bd4e36c402/README.md

PR #51, opened 2022-03-28
feat: Bump confluent kafka to 1.6.1

lynnagara:
Updates confluent-kafka (and librdkafka) from 1.5.0 to 1.6.1.
This upgrade includes support for incremental consumer rebalancing
which we would like to try.


volokluev (approved):
Nice one

PR #53, opened 2022-03-30
feat(consumer): Support incremental cooperative rebalancing

lynnagara:
KafkaConsumer now supports the `cooperative-sticky` partitioning strategy.
The goal of this change is to try and reduce the amount of unnecessary rebalances
that happen during Kubernetes rolling deployments.

### DIFF
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -252,63 +252,63 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         ) -> None:
             self.__state = KafkaConsumerState.ASSIGNING
 
-            try:
-                assignment: MutableSequence[ConfluentTopicPartition] = []
+            if self.__incremental is True:
+                incremental_assignment: MutableSequence[ConfluentTopicPartition] = []
 
-                for partition in self.__consumer.committed(partitions):
+                for partition in partitions:
                     if partition.offset >= 0:
-                        assignment.append(partition)
+                        incremental_assignment.append(partition)
                     elif partition.offset == OFFSET_INVALID:
-                        assignment.append(
+                        incremental_assignment.append(
                             self.__resolve_partition_starting_offset(partition)
                         )
                     else:
                         raise ValueError("received unexpected offset")
 
-                offsets: MutableMapping[Partition, int] = {
-                    Partition(Topic(i.topic), i.partition): i.offset for i in assignment
-                }
+                self.__incremental_assign(
+                    {
+                        Partition(Topic(i.topic), i.partition): i.offset
+                        for i in incremental_assignment
+                    }
+                )
 
-                if self.__incremental is True:
-                    incremental_assignment: MutableSequence[
-                        ConfluentTopicPartition
-                    ] = []
+            else:
+                try:
+                    assignment: MutableSequence[ConfluentTopicPartition] = []
 
-                    for partition in partitions:
+                    for partition in self.__consumer.committed(partitions):
                         if partition.offset >= 0:
-                            incremental_assignment.append(partition)
+                            assignment.append(partition)
                         elif partition.offset == OFFSET_INVALID:
-                            incremental_assignment.append(
+                            assignment.append(
                                 self.__resolve_partition_starting_offset(partition)
                             )
                         else:
                             raise ValueError("received unexpected offset")
 
-                    self.__incremental_assign(
-                        {
-                            Partition(Topic(i.topic), i.partition): i.offset
-                            for i in assignment
-                        }
-                    )
-                else:
+                    offsets: MutableMapping[Partition, int] = {
+                        Partition(Topic(i.topic), i.partition): i.offset
+                        for i in assignment
+                    }
+
                     self.__assign(offsets)
 
-                # Ensure that all partitions are resumed on assignment to avoid
-                # carrying over state from a previous assignment.
-                self.__consumer.resume(
-                    [
-                        ConfluentTopicPartition(
-                            partition.topic.name, partition.index, offset
-                        )
-                        for partition, offset in offsets.items()
-                    ]
-                )
+                    # Ensure that all partitions are resumed on assignment to avoid
+                    # carrying over state from a previous assignment.
+                    self.__consumer.resume(
+                        [
+                            ConfluentTopicPartition(
+                                partition.topic.name, partition.index, offset
+                            )
+                            for partition, offset in offsets.items()
+                        ]
+                    )
 
-                for partition in offsets:
-                    self.__paused.discard(partition)
-            except Exception:
-                self.__state = KafkaConsumerState.ERROR
-                raise
+                    for partition in offsets:
+                        self.__paused.discard(partition)
+                except Exception:
+                    self.__state = KafkaConsumerState.ERROR
+                    raise
 
             try:
                 if on_assign is not None:
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -146,7 +146,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         configuration: MutableMapping[str, Any],
         *,
         commit_retry_policy: Optional[RetryPolicy] = None,
-        incremental: bool = False,
+        incremental_cooperative: bool = False,
     ) -> None:
         if commit_retry_policy is None:
             commit_retry_policy = NoRetryPolicy()
@@ -183,9 +183,9 @@ class KafkaConsumer(Consumer[KafkaPayload]):
                 "invalid value for 'enable.auto.offset.store' configuration"
             )
 
-        self.__incremental = incremental
+        self.__incremental_cooperative = incremental_cooperative
 
-        if self.__incremental is True:
+        if self.__incremental_cooperative is True:
             configuration["partition.assignment.strategy"] = "cooperative-sticky"
 
         # NOTE: Offsets are explicitly managed as part of the assignment
@@ -252,7 +252,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         ) -> None:
             self.__state = KafkaConsumerState.ASSIGNING
 
-            if self.__incremental is True:
+            if self.__incremental_cooperative is True:
                 incremental_assignment: MutableSequence[ConfluentTopicPartition] = []
 
                 for partition in partitions:
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -106,7 +106,8 @@ def as_kafka_configuration_bool(value: Any) -> bool:
 
 class KafkaConsumer(Consumer[KafkaPayload]):
     """
-    The behavior of this consumer differs slightly from the Confluent
+    If a non-cooperative partition assignment strategy is selected,
+    the behavior of this consumer differs slightly from the Confluent
     consumer during rebalancing operations. Whenever a partition is assigned
     to this consumer, offsets are *always* automatically reset to the
     committed offset for that partition (or if no offsets have been committed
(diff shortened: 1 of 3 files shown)


arroyo/backends/kafka/consumer.py
```diff
@@ -246,40 +252,63 @@ def assignment_callback(
         ) -> None:
             self.__state = KafkaConsumerState.ASSIGNING
 
-            try:
-                assignment: MutableSequence[ConfluentTopicPartition] = []
+            if self.__incremental_cooperative is True:
```
lynnagara, line 255: This diff got pretty weird because of indentation. This block from here to line 273 is new though, and could use a close review. It's quite different to what happens with eager rebalancing (lines 276-305), which is essentially unchanged.

fpacifici (commented):
Have you tried to test locally how this performs?
Specifically:
- how long it takes to assign a new partition to a new consumer showing up 
- how long it takes to revoke all remaining partitions (if any) from a consumer that is leaving the group? Does the group wait `scheduled.rebalance.max.delay.ms` before reassigning the partitions to a new consumer?
- how long does it take to have partitions reassigned in case a node leaves and rejoins? 

I think we should do a test locally before doing it in prod.

arroyo/backends/kafka/consumer.py
```diff
@@ -182,6 +183,11 @@ def __init__(
                 "invalid value for 'enable.auto.offset.store' configuration"
             )
 
+        self.__incremental_cooperative = incremental_cooperative
+
+        if self.__incremental_cooperative is True:
+            configuration["partition.assignment.strategy"] = "cooperative-sticky"
```
fpacifici, line 189: Why providing `incremental_cooperative` as a dedicated flag instead of passing it through the `configuration` mapping like all the others config parameters ?

arroyo/backends/kafka/consumer.py
```diff
@@ -431,29 +471,33 @@ def __validate_offsets(self, offsets: Mapping[Partition, int]) -> None:
...
-            self.__consumer.assign(
-                [
-                    ConfluentTopicPartition(
-                        partition.topic.name, partition.index, offset
-                    )
-                    for partition, offset in offsets.items()
-                ]
+        for partition, offset in offsets.items():
+            self.__consumer.seek(
+                ConfluentTopicPartition(partition.topic.name, partition.index, offset)
```
fpacifici, line 499: Why this change ?

arroyo/backends/kafka/consumer.py
lynnagara: I had it the other way originally but figured it was easier to pass a boolean than remember this string then check for it to decide whether to apply incremental_assign. Thinking about it again, I might switch back to avoid changing the interface of this class though.

arroyo/backends/kafka/consumer.py
```diff
@@ -246,40 +252,74 @@ def assignment_callback(
...
-                        for partition, offset in offsets.items()
-                    ]
-                )
+            if self.__incremental_cooperative is True:
+                try:
+                    incremental_assignment: MutableSequence[
+                        ConfluentTopicPartition
+                    ] = []
+
+                    for partition in partitions:
```
fpacifici, line 261: If I am getting this right, you are not asking the broker anymore for the last committed offset. 
Which means that you are invoking `incremental_assign` with the offset you received in the `assignment_callback` call. 
Is this intentional? In the stop-the-world rebalancing, fetching the last committed offset and seeking that was meant to ensure all partitions had the same behavior (whether they were previously assigned to the same consumer or not). All partitions would be reset to the last committed offset.
I would expect it to be impossible, in the incremental case, to receive a partition you were already owning before (unless the consumer restarted). Is this guaranteed or can you still be assigned a partition you already owned before rebalancing?
If it is guaranteed and all partitions you received are "new" to this consumer, are you not resetting the offset because the offset you receive are supposed to already be the last committed ones ?

arroyo/backends/kafka/consumer.py
lynnagara: I found it pretty unintuitive that the same `__seek` implementation was being called from both `seek()` and the assign callback then checking the KafkaConsumerState to effectively determine which one was making the call. Now assigning function is separate and never calls this `__seek` function anymore.

arroyo/backends/kafka/consumer.py
```diff
@@ -246,40 +252,74 @@ def assignment_callback(
...
+                        [
+                            ConfluentTopicPartition(
+                                partition.topic.name, partition.index, offset
+                            )
+                            for partition, offset in offsets.items()
+                        ]
+                    )
+
+                    for partition in offsets:
+                        self.__paused.discard(partition)
```
fpacifici, line 319: Any reason for relying on the `resume` method for the incremental method and, instead, reimplementing its logic here for the stop-the-world rebalancing ?

arroyo/backends/kafka/consumer.py
fpacifici: Also, now this method is only called by the `seek` public method. Why not inlining it there ?

tests/backends/test_kafka.py
```diff
@@ -133,6 +144,70 @@ def test_auto_offset_reset_error(self) -> None:
...
+        {
+            **configuration,
+            "auto.offset.reset": "earliest",
+            "enable.auto.commit": False,
+            "enable.auto.offset.store": False,
+            "group.id": group_id,
+            "session.timeout.ms": 10000,
+        },
+        incremental_cooperative=True,
+    )
```
fpacifici, line 178: Thanks.
By the way, what happens if a number of consumers are incremental and others are not ? Which means what will happen at the first deployment ?

tests/backends/test_kafka.py
```diff
@@ -133,6 +144,70 @@ def test_auto_offset_reset_error(self) -> None:
...
+        producer
+    ), closing(consumer_a), closing(consumer_b):
+        for i in range(10):
+            for j in range(partitions_count):
+                producer.produce(
+                    Partition(topic, j),
+                    KafkaPayload(None, f"{j}-{i}".encode("utf8"), []),
+                )
+
+        consumer_a.subscribe([topic])
```
fpacifici, line 190: Why not providing callbacks so that you can assert that only one partition is transferred ?

arroyo/backends/kafka/consumer.py
fpacifici: > I might switch back to avoid changing the interface of this class though.

Also we would not reintroduce two separate ways to provide Kafka config that took us long to clean up

tests/backends/test_kafka.py
lynnagara: I guess that could've worked too. Still this works well enough in checking the same thing and it's copied from the test of the eager consumer.

arroyo/backends/kafka/consumer.py
lynnagara: No reason, I just mostly avoided touching this implementation (it's just indented differently so it shows up here). But yes it's cleaner to reuse `resume` everywhere, I'll switch it over.

arroyo/backends/kafka/consumer.py
lynnagara: I was avoiding touching a lot of code earlier, but yes it is better inline since no one else calls it now. Updated.

tests/backends/test_kafka.py
fpacifici: Right but there is no way to tell whether all partitions were revoked and then reassigned or whether the incremental assignment actually works.

lynnagara: Yeah this was not about test Kafka consumer internals, just that we end up with one partition assigned to each consumer and consuming on both works ok.

arroyo/backends/kafka/consumer.py
lynnagara: I don't *think* you can receive a partition you already owned, but in any case I don't quite get why we would need to check the offset since the offset you are receiving should already be the correct one. The other issue I noticed with `committed()` is that it hangs forever (or times out if you pass it a timeout) for the cooperative-sticky strategy in the scenario where it's a new consumer group and no offset has been committed yet.

fpacifici: This will be important to verify. 
The reason for getting the `committed()` offsets is so that all partitions start in a consistent way from the last committed offset after a rebalance (without doing so the partitions that are reassigned to you would start from the last consumed offset). 
This behavior is visible to the application, but if we commit during revoke we should not have any issue as the two behaviors would be identical. 
If we never receive a partition we already owned and we consistently commit on revoke we are good. But if we fail in committing during revoke and we get the same partition we owned in the past I don't know what would happen if we do not reset the offsets. 
If `committed()` is taking much longer, then it is probably a non starter. Could you please verify that even if we did receive the same partition again on subscriptions (where we are testing this), we will not risk missing messages ?

lynnagara: > The reason for getting the `committed()` offsets is so that all partitions start in a consistent way from the last committed offset after a rebalance (without doing so the partitions that are reassigned to you would start from the last consumed offset).

I don't think this is true. The offsets passed to the assignment callback are already the committed ones not just the ones that are consumed.

Kafka guarantees that
- only new partitions are passed in this callback not previously owned ones
- The revoke callback is always triggered before the assignment one.

The reason for committing offsets during revoke is to avoid double processing them so the new consumer doesn't get that same offset again.

lynnagara:
> * how long it takes to assign a new partition to a new consumer showing up
> * how long it takes to revoke all remaining partitions (if any) from a consumer that is leaving the group? Does the group wait `scheduled.rebalance.max.delay.ms` before reassigning the partitions to a new consumer?
> * how long does it take to have partitions reassigned in case a node leaves and rejoins?

It's typically not as long as `scheduled.rebalance.max.delay.ms` which is 5 minutes by default. All of these usually take around 3 seconds in my local testing with 2 consumers. The test case I added adds a new consumer to an existing group and waits for a partition to assigned to it. We wait for up to 10 seconds max and it generally always passes.

arroyo/backends/kafka/consumer.py
fpacifici: > The offsets passed to the assignment callback are already the committed ones not just the ones that are consumed.

Are you sure. I thought this was the reason for making this consumer behave differently from the standard Kafka consumer https://github.com/getsentry/arroyo/blob/main/arroyo/backends/kafka/consumer.py#L109-L118.

Anyway, if Kafka guarantees that only new partitions are passed and not previously owned one, we should be good.

tests/backends/test_kafka.py
fpacifici: Ok, let's do it one day I am on PTO. 
Out of curiosity, what happens if you start a consumer that works with the incremental rebalancing system to a group that is using the standard system ?

arroyo/backends/kafka/consumer.py
fpacifici: I am talking about all those fields we were passing via CLI that we are slowly moving to settings.

fpacifici (approved):
Seems good to try

fpacifici:
I think you should update the docstring of the KafkaConsumer class so that it reflects the behavior on incremental cooperative rebalancing. Specifically this https://github.com/getsentry/arroyo/blob/main/arroyo/backends/kafka/consumer.py#L109-L118

arroyo/backends/kafka/consumer.py
lynnagara: Yes, but this only applies to the partitions that are being newly assigned to the consumer. Previously all of the partitions (including those already assigned to the consumer) would be provided in the callback, now only the incremental ones are. So we are no longer rewinding back those partitions which the consumer is maintaining across a rebalance like we use to, those just continue from the same place.

arroyo/backends/kafka/consumer.py
lynnagara: Unfortunately passing via CLI is still the easiest option currently as we don't have a mechanism yet to provide different settings for consumers/producers of a topic via settings.

mwarkentin:
@lynnagara 🎉 

Let me know if / when you need some help testing this. :)

PR #58, opened 2022-04-06
feat: Support incremental assignments in stream processor

lynnagara:
With the incremental assign / revoke strategy the assignment callback can
be triggered without a corresponding revocation beforehand. In that case
the processing strategy will still be active. When this happens, we close
the strategy before recreating it to make the behavior compatible with
the eager strategy.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -71,19 +71,29 @@ class StreamProcessor(Generic[TPayload]):
             self.__processing_strategy = None
             self.__message = None  # avoid leaking buffered messages across assignments
 
-        def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
-            if self.__processing_strategy is not None:
-                _close_strategy()
-
-            logger.info("New partitions assigned: %r", partitions)
+        def _create_strategy() -> None:
             self.__processing_strategy = self.__processor_factory.create(self.__commit)
             logger.debug(
                 "Initialized processing strategy: %r", self.__processing_strategy
             )
 
+        def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
+            logger.info("New partitions assigned: %r", partitions)
+            if self.__processing_strategy is not None:
+                _close_strategy()
+            _create_strategy()
+
         def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
             logger.info("Partitions revoked: %r", partitions)
             _close_strategy()
+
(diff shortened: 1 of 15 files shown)


fpacifici (approved):
This works because the only case where you can receive new assigned partitions without prior revocation is an incremental rebalancing and the partitions you receive were not assigned before and you are not resetting any offsets (when you call `assign_incremental` you are assigning the same offset that is the committed offset). Is that correct?

Now the question is: what if you loose partitions without gaining new ones? you will have a call to `revoke` and that will close the strategy, but that will not be followed by an assignment so you will remain without a strategy instantiated. How do you plan to deal with that ?

lynnagara:
@fpacifici Made a few changes since you last reviewed to handle incremental revoke. Now we check on revocation if the consumer has any other assigned partitions and if so the strategy is also recreated. Could you take another look at let me know if this makes sense to you?

arroyo/processing/processor.py
```diff
@@ -85,6 +71,30 @@ def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
...
+            logger.info("Partitions revoked: %r", partitions)
+            _close_strategy()
+
+            # Recreate the strategy if the consumer still has other partitions
+            # assigned and is not closed or errored
+            try:
+                if self.__consumer.tell().keys() - set(partitions):
+                    _create_strategy()
+            except Exception:
+                pass
```
fpacifici, line 96: Should we crash instead? If you cannot call `tell` we would not be creating the strategy and we would run into an inconsistent state.

fpacifici (approved):
Please see my comment. Otherwise it makes sense.

Any chance you can make separate tests for the cases where all partitions are revoked vs only some of them ?

arroyo/processing/processor.py
lynnagara: Yeah, this was mostly just to catch consumer closed / terminated exception. I can just catch those specific errors instead and let anything else crash.

lynnagara: Actually on second thoughts, I think I would leave this as-is. it we cannot call `tell` it only means that the consumer is in the process of shutting down and we should not be re-creating the strategy otherwise it'll mess with the shutdown sequence.

fpacifici: log a warning ?

lynnagara: Ended up changing this line to catch RuntimeError only. So now it will crash on anything else including networking issues.

PR #61, opened 2022-04-20
fix(dlq): Updated DLQ logic for ParallelTransformStep

rahul-kumar-saini:
### Overview
- We want to be able to batch bad messages together and throw them at once so entire batch is not failed on single bad message
- Current implementation attempts to support this, but lacks fundamental integration with the ParallelTransformStep which is the main batching process

### ParallelTransformStep

**Here is my understanding of how this works:**
- A message is submitted via `submit()` method, this message is added to a batch
- If this batch is filled, the batch is submitted for processing via `__submit_batch()`
- `__submit_batch()` launches parallel processing of all messages (the transform step aka message processor)
- These processes are checked up on via `poll()` which calls `__check_for_results()`
- `__check_for_results()` gets back all the results of done processes
- The messages outputted by the transform steps are then submitted to the next processing step

**How do we integrate DLQ logic here?**
- Both the transform step (message processor) itself and the next processing step can raise encounter a bad message and in turn raise an exception for it
- Therefore, both the `get()` call on the async results and the `submit()` call to the next step should be surrounded by a try/catch in order to continue processing the batch without interruption
- These try/catch's should gather the exceptions into a list
- Once we are satisfied that the good messages are all through, raise a single exception containing the group of exceptions collected
    - This should happen at the end of `__check_for_results()`

### Changes
- Updated ParallelTransformStep to gather bad messages and throw them at once
- Replaced `InvalidMessages` with `InvalidMessage` and `InvalidBatchedMessages` exceptions
- Workers would generally only raise `InvalidMessage`, the batched version is strictly a collection of the singular version and would only be raised by batching processes

### DIFF
--- a/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
+++ b/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
@@ -1,15 +1,18 @@
 import logging
-from typing import Optional
+from typing import Optional, Union
 
 from arroyo.processing.strategies.abstract import ProcessingStrategy as ProcessingStep
 from arroyo.processing.strategies.dead_letter_queue.policies.abstract import (
     DeadLetterQueuePolicy,
-    InvalidMessages,
+    InvalidBatchedMessages,
+    InvalidMessage,
 )
 from arroyo.types import Message, TPayload
 from arroyo.utils.metrics import get_metrics
 
 logger = logging.getLogger(__name__)
+
+RECEIVED_MESSAGE_METRIC = "dlq.received_message"
 
 
 class DeadLetterQueue(ProcessingStep[TPayload]):
@@ -17,7 +20,8 @@ class DeadLetterQueue(ProcessingStep[TPayload]):
     DLQ Processing Step.
 
     Attempts to submit a given message to the next processing step,
-    handling `InvalidMessages` according to the given Policy.
+    handling `InvalidMessage` and `InvalidBatchedMessages`
+    according to the given Policy.
     """
 
     def __init__(
@@ -33,17 +37,23 @@ class DeadLetterQueue(ProcessingStep[TPayload]):
     def poll(self) -> None:
         try:
             self.__next_step.poll()
-        except InvalidMessages as e:
-            self.__metrics.increment("dlq.received_message", len(e.messages))
-            self.__policy.handle_invalid_messages(e)
+        except (InvalidMessage, InvalidBatchedMessages) as e:
+            self._handle_invalid_messages(e)
 
     def submit(self, message: Message[TPayload]) -> None:
         assert not self.__closed
         try:
             self.__next_step.submit(message)
-        except InvalidMessages as e:
-            self.__metrics.increment("dlq.received_message", len(e.messages))
-            self.__policy.handle_invalid_messages(e)
+        except (InvalidMessage, InvalidBatchedMessages) as e:
+            self._handle_invalid_messages(e)
+
+    def _handle_invalid_messages(
+        self, e: Union[InvalidMessage, InvalidBatchedMessages]
+    ) -> None:
+        if isinstance(e, InvalidMessage):
+            e = InvalidBatchedMessages([e])
+        self.__metrics.increment(RECEIVED_MESSAGE_METRIC, len(e.exceptions))
+        self.__policy.handle_invalid_messages(e)
 
     def close(self) -> None:
         self.__closed = True
--- a/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
+++ b/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
@@ -58,5 +58,6 @@ class DeadLetterQueue(ProcessingStep[TPayload]):
         self.__next_step.terminate()
 
     def join(self, timeout: Optional[float] = None) -> None:
+        self.__policy.join(timeout)
         self.__next_step.close()
         self.__next_step.join(timeout)
--- a/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
+++ b/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
@@ -19,8 +19,7 @@ class DeadLetterQueue(ProcessingStep[TPayload]):
     DLQ Processing Step.
 
     Attempts to submit a given message to the next processing step,
-    handling `InvalidMessage` and `InvalidBatchedMessages`
-    according to the given Policy.
+    handling `InvalidMessages` according to the given Policy.
     """
 
     def __init__(
--- a/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
+++ b/arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
@@ -1,11 +1,10 @@
 import logging
-from typing import Optional, Union
+from typing import Optional
 
 from arroyo.processing.strategies.abstract import ProcessingStrategy as ProcessingStep
 from arroyo.processing.strategies.dead_letter_queue.policies.abstract import (
     DeadLetterQueuePolicy,
-    InvalidBatchedMessages,
-    InvalidMessage,
+    InvalidMessages,
 )
 from arroyo.types import Message, TPayload
 from arroyo.utils.metrics import get_metrics
@@ -37,22 +36,18 @@ class DeadLetterQueue(ProcessingStep[TPayload]):
     def poll(self) -> None:
         try:
             self.__next_step.poll()
-        except (InvalidMessage, InvalidBatchedMessages) as e:
+        except InvalidMessages as e:
             self._handle_invalid_messages(e)
 
     def submit(self, message: Message[TPayload]) -> None:
         assert not self.__closed
         try:
             self.__next_step.submit(message)
-        except (InvalidMessage, InvalidBatchedMessages) as e:
+        except InvalidMessages as e:
             self._handle_invalid_messages(e)
 
-    def _handle_invalid_messages(
-        self, e: Union[InvalidMessage, InvalidBatchedMessages]
-    ) -> None:
-        if isinstance(e, InvalidMessage):
-            e = InvalidBatchedMessages([e])
-        self.__metrics.increment(RECEIVED_MESSAGE_METRIC, len(e.exceptions))
+    def _handle_invalid_messages(self, e: InvalidMessages) -> None:
+        self.__metrics.increment(RECEIVED_MESSAGE_METRIC, len(e.messages))
         self.__policy.handle_invalid_messages(e)
 
     def close(self) -> None:
--- a/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
+++ b/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
@@ -1,18 +1,49 @@
 from abc import ABC, abstractmethod
-from typing import Any, Sequence
+from datetime import datetime
+from typing import Optional, Sequence, Union
+
+Serializable = Union[str, bytes]
 
 
-class InvalidMessages(Exception):
+class InvalidMessage(Exception):
     """
-    An exception to be thrown to pass bad messages to the DLQ
-    so they are handled correctly.
+    An exception to be thrown to pass a bad message to the DLQ.
+
+    A reason can be optionally passed describing why the message
+    is bad.
+
+    If the original topic the message was produced to is known,
+    that should be passed along via this exception.
     """
 
-    def __init__(self, messages: Sequence[Any]):
-        self.messages = messages
+    def __init__(
+        self,
+        message: Serializable,
+        reason: Optional[str] = None,
+        original_topic: Optional[str] = None,
+    ):
+        self.message = message
+        self.reason = reason or "unknown"
+        self.topic = original_topic or "unknown"
+        self.timestamp = str(datetime.now())
 
     def __str__(self) -> str:
-        return f"Invalid Message(s): {self.messages}"
+        return (
+            f"Invalid Message originally produced to: {self.topic}\n"
+            f"Reason: {self.reason}\n"
+            f"Exception thrown at: {self.timestamp}\n"
+            f"Message: {str(self.message)}"
+        )
+
+
+class InvalidBatchedMessages(Exception):
+    """
+    WIP
+    Should only be used by batching strategies.
+    """
+
+    def __init__(self, exceptions: Sequence[InvalidMessage]):
+        self.exceptions = exceptions
 
 
 class DeadLetterQueuePolicy(ABC):
@@ -21,8 +52,8 @@ class DeadLetterQueuePolicy(ABC):
     """
 
     @abstractmethod
-    def handle_invalid_messages(self, e: InvalidMessages) -> None:
+    def handle_invalid_messages(self, e: InvalidBatchedMessages) -> None:
         """
-        Decide what to do with invalid messages.
+        Decide what to do with a batch of invalid messages.
         """
         pass
--- a/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
+++ b/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
@@ -38,8 +38,9 @@ class InvalidMessage(Exception):
 
 class InvalidBatchedMessages(Exception):
     """
-    WIP
-    Should only be used by batching strategies.
+    An exception to be thrown to pass batched invalid messages.
+    to the DLQ. Generally should only be used by batching
+    strategies which collect multiple `InvalidMessage` exceptions.
     """
 
     def __init__(self, exceptions: Sequence[InvalidMessage]):
--- a/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
+++ b/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
@@ -1,19 +1,95 @@
+import base64
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Optional, Sequence, Union
+from typing import Mapping, Optional, Sequence, Tuple, Union
 
-Serializable = Union[str, bytes]
+from arroyo.backends.kafka.consumer import Headers
+
+SerializedPayload = Union[str, bytes]
+DeserializedHeaders = Sequence[Tuple[str, str]]
+JSONSerializable = Optional[Union[str, int, datetime, DeserializedHeaders]]
+
+
+class InvalidMessage(ABC):
+    """
+    A class representing a bad message to be passed to the DLQ.
+
+    An `InvalidMessages` exception should be raised containing
+    one or more `InvalidMessage` objects in order to actually
+    pass data to the DLQ.
+
+    If a produce policy is configured on the relevant DLQ, a
+    message in the form returned by `to_dict()` will be produced
+    via the policy.
+    """
+
+    @abstractmethod
+    def to_dict(self) -> Mapping[str, JSONSerializable]:
+        raise NotImplementedError
+
+    def _deserialize_payload(self, payload: SerializedPayload) -> str:
+        if isinstance(payload, bytes):
+            return self._deserialize_bytes(payload)
+        return payload
+
+    def _deserialize_bytes(self, value: bytes) -> str:
+        try:
+            decoded = value.decode("utf-8")
+        except UnicodeDecodeError:
+            decoded = "(base64) " + base64.b64encode(value).decode("utf-8")
+        return decoded
 
 
 @dataclass(frozen=True)
-class InvalidMessage:
-    payload: Serializable
-    timestamp: datetime
+class InvalidRawMessage(InvalidMessage):
+    """
+    A dataclass to generally represent any kind of bad message.
+    A `reason` can be provided for debugging purposes.
+    """
+
+    payload: SerializedPayload
     reason: Optional[str] = None
-    consumer_group: Optional[str] = None
-    partition: Optional[int] = None
-    offset: Optional[int] = None
+
+    def to_dict(self) -> Mapping[str, JSONSerializable]:
+        return {"payload": super()._deserialize_payload(self.payload)}
+
+
+@dataclass(frozen=True)
+class InvalidKafkaMessage(InvalidMessage):
+    """
+    A dataclass to generally represent a bad Kafka message.
+    A `reason` can be provided for debugging purposes.
+    """
+
+    payload: SerializedPayload
+    timestamp: datetime
+    topic: str
+    consumer_group: str
+    partition: int
+    offset: int
+    headers: Headers
+    key: Optional[bytes] = None
+    reason: Optional[str] = None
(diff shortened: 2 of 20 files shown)


arroyo/processing/strategies/dead_letter_queue/dead_letter_queue.py
```diff
@@ -33,17 +37,23 @@ def __init__(
...
+        except (InvalidMessage, InvalidBatchedMessages) as e:
+            self._handle_invalid_messages(e)
+
+    def _handle_invalid_messages(
+        self, e: Union[InvalidMessage, InvalidBatchedMessages]
+    ) -> None:
+        if isinstance(e, InvalidMessage):
+            e = InvalidBatchedMessages([e])
+        self.__metrics.increment(RECEIVED_MESSAGE_METRIC, len(e.exceptions))
+        self.__policy.handle_invalid_messages(e)
```
rahul-kumar-saini, line 56: With this implementation, a policy decides how to deal with multiple bad messages. If I had just looped and submitted bad messages one by one to a policy, we would lose part of the batch of bad messages if the policy decides to raise an exception part way through.

This would make our count policy have an inaccurate count since it would not have counted up all bad messages before throwing the exception.

tests/processing/strategies/test_streaming.py
```diff
@@ -405,6 +408,48 @@ def test_parallel_transform_worker_apply() -> None:
...
+    index, output_batch, bad_messages = parallel_transform_worker_apply(
+        fail_bad_messages,
+        input_batch,
+        output_block,
+    )
+    # all 10 messages processed
+    assert index == 10
+    # 5 were bad, 5 were good
+    assert len(bad_messages) == 5
+    assert len(output_batch) == 5
```
rahul-kumar-saini, line 449: Not sure if the length of output batch has other logical implications, but out of 10 messages, 5 were processed successfully and 5 were raised together to DLQ.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -225,12 +231,40 @@ def parallel_transform_worker_initializer(
...
+    bad_messages: MutableSequence[InvalidMessage]
+
+
 def parallel_transform_worker_apply(
     function: Callable[[Message[TPayload]], TTransformed],
     input_batch: MessageBatch[TPayload],
     output_block: SharedMemory,
     start_index: int = 0,
-) -> Tuple[int, MessageBatch[TTransformed]]:
+    bad_messages: MutableSequence[InvalidMessage] = [],
```
fpacifici, line 266: Please do not pass a mutable object to a function to simulate a different return type.
Besides making the code extremely hard to understand, this list will never be reset if you use the default parameters.
If you need to return a different type than what you returned before, change the return type.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -239,9 +273,13 @@ def parallel_transform_worker_apply(
 
         try:
             result = function(message)
+        except InvalidMessage as e:
+            bad_messages.append(e)
```
fpacifici, line 277: Have you tested this on more than one process to ensure that the `InvalidMessage` exception actually pickles properly to be transferred from a subprocess to the main one ?

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -354,27 +391,31 @@ def __submit_batch(self) -> None:
...
 
         # TODO: This does not handle rejections from the next step!
-        for message in output_batch:
-            self.__next_step.poll()
-            self.__next_step.submit(message)
+        for message in result.output_batch:
+            try:
+                self.__next_step.poll()
+                self.__next_step.submit(message)
+            except InvalidMessage as e:
```
fpacifici, line 412: I don't think this is correct. the `poll` could be called on a batching step itself that can return multiple bad messages.

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
```diff
@@ -1,18 +1,50 @@
...
+
+class InvalidBatchedMessages(Exception):
+    """
+    An exception to be thrown to pass batched invalid messages.
+    to the DLQ. Generally should only be used by batching
+    strategies which collect multiple `InvalidMessage` exceptions.
+    """
+
+    def __init__(self, exceptions: Sequence[InvalidMessage]):
+        self.exceptions = exceptions
```
fpacifici, line 47: I don't see the reason for this exception. I think your approach in PR https://github.com/getsentry/arroyo/pull/57 is better. Have one exception and that can contain multiple payloads, one per bad message. 
I don't see a reason for representing an individual bad message as an exception itself.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -225,12 +231,40 @@ def parallel_transform_worker_initializer(
...
+    should be passed to a DLQ by raising the
+    `InvalidBatchedMessages` exception.
+    """
+
+    index_processed_til: int
+    output_batch: MessageBatch[TTransformed]
+    bad_messages: MutableSequence[InvalidMessage]
+
+
 def parallel_transform_worker_apply(
```
fpacifici, line 261: Please have a unit test for this function with the changes you added.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -268,8 +306,7 @@ def parallel_transform_worker_apply(
                 break
         else:
             i += 1
-
-    return (i, output_batch)
+    return ParallelTransformResult(i, output_batch, bad_messages)
 
 
 class ParallelTransformStep(ProcessingStep[TPayload]):
```
fpacifici, line 312: Please have a unit test for this class with the changes you added. There should be already a unit test that tests this class in isolation.

fpacifici (changes requested):
I think the approach (let the individual process process each message and return a list of successful and broken message is good. 
Please see comments in line.

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: I'm not sure what you mean, afaik `MutableSequence` is just a regular `Sequence` but refers to the list itself being mutable so `mypy` doesn't complain when I use the `.append()` method of the class. The input type and return type are still the same, it's a list of `InvalidMessage`.

The point of this was to be able to pass the partial list of bad messages back into this function in case the batch wasn't fully processed, just like we pass a `start_index` back in when rerunning on the same batch part way through.

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: I've already added a test here which tests the case of some bad and some good messages:
https://github.com/getsentry/arroyo/blob/2535e2f7391e35493e04e6c14045b2e0ebb0c737/tests/processing/strategies/test_streaming.py#L417

Is there another case you'd like tested?

Also see my comment:
https://github.com/getsentry/arroyo/pull/61#discussion_r854545804
I'm uncertain if this ok^

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
rahul-kumar-saini: I'll refactor this (or maybe merge that in here?) to follow the same approach in that PR once that is finalized.

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: Done, I commented what I think is going on with the parallel processes asserts

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: I've now added a test to ensure multiple processes + pickle works

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: Added another except for this just below, but I'll refactor following this:
https://github.com/getsentry/arroyo/pull/61#discussion_r855610674

tests/processing/strategies/test_streaming.py
```diff
@@ -505,3 +565,58 @@ def test_parallel_transform_step_terminate_workers() -> None:
         starting_processes,
     ), assert_changes(lambda: int(next_step.terminate.call_count), 0, 1):
         transform_step.terminate()
+
+
+def test_parallel_transform_step_bad_messages() -> None:
```
rahul-kumar-saini, line 570: I couldn't figure out why, but moving this test on top of `test_parallel_transform_step` breaks that test locally (it raises a `ChildProcessTerminated` exception). I didn't test on CI but maybe someone has an idea as to why that could be?

I'm guessing there's some cleanup I'm not doing that affects the next test but I'm not sure what it could be.

tests/processing/strategies/test_streaming.py
```diff
@@ -505,3 +564,58 @@ def test_parallel_transform_step_terminate_workers() -> None:
...
+        # submit 9 messages: 4 good ones 5 bad ones
+        for message in messages:
+            transform_step.submit(message)
+            transform_step.poll()
+
+        # wait for all processes to finish
+        with pytest.raises(InvalidMessages) as e_info:
+            transform_step.join()
+
+        transform_step.close()
```
lynnagara, line 613: close() should be called before join()

arroyo/processing/strategies/streaming/transform.py
lynnagara: Why do you need to mutate the array that is passed in though? And why does it need to be passed at all - shouldn't it always be empty initially anyway?

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -239,9 +273,13 @@ def parallel_transform_worker_apply(
 
         try:
             result = function(message)
+        except InvalidMessages as e:
+            bad_messages += e.messages
+            i += 1
```
lynnagara, line 278: The way this loop is written is a bit unintuitive to me. Is there any particular reason we can't just write:
```
i = 0
for message in input_batch:
    ...
```

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: Same reason `start_index` exists, if the input batch is partially processed, then next iteration an index is passed back in to know where to start processing again from.

So if the input batch is partially processed, then the array of bad messages is potentially not complete yet either and must be passed back in to continue being filled in case of bad messages in the rest of the batch.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -225,12 +231,40 @@ def parallel_transform_worker_initializer(
         custom_initialize_func()
 
 
+@dataclass
+class ParallelTransformResult(Generic[TTransformed]):
+    """
+    Results of applying the transform function to a batch
+    of messages in parallel.
+
+    `index_processed_til` holds the index of the next
```
lynnagara, line 240: nit: `index_processed_until` vs `index_processed_til`. Actually, `until` has an inclusive connotation to me (as in message was already processed) but it actually hasn't been.

Though I'm not really sure if it's really needed at all? This implementation would be simpler without it. What kind of scenario would actually cause this to happen?

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: Function is called again with the index to start processing again from when `ValueTooLarge` is caught I believe.

https://github.com/getsentry/arroyo/blob/9b801a2892d6eb39367768482f102a38f5171de8/arroyo/processing/strategies/streaming/transform.py#L415-L437

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: I'm not sure what to call it, it was just called `i` before and I wanted to make it more clear what this magic index meant. Maybe something like `next_index_to_process`?

Explanation in prev comment.

lynnagara: Ah ok. It's called `start_index` in the parallel transform function

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
```diff
@@ -1,5 +1,19 @@
...
+
+
+@dataclass(frozen=True)
+class InvalidMessage:
+    payload: Serializable
+    timestamp: datetime
+    reason: Optional[str] = None
+    consumer_group: Optional[str] = None
+    partition: Optional[int] = None
+    offset: Optional[int] = None
```
lynnagara, line 16: No Kafka topic? I actually wonder if we should just remove all this and have the user of this library put it into the payload instead. Otherwise I feel like we will be changing this list too frequently as it is quite rigid and cannot account for everyone's usecase.

rahul-kumar-saini: The reason we dropped the Kafka topic is cause we decided each (consumer? storage?) will have its own dead letter topic. We don't need topic info here anymore since we know which one it is from which DLQ topic a message is produced to.

I also agree that we could just let the user decide entirely how the message should look, which is mostly why these are optional anyway.

lynnagara: My question is really whether we should just remove all of these though? The inclusion of these particular items (and not topic for example) seems *extremely* specific to our usage

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
```diff
@@ -1,5 +1,19 @@
...
-from typing import Any, Sequence
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Optional, Sequence, Union
+
+Serializable = Union[str, bytes]
+
+
+@dataclass(frozen=True)
+class InvalidMessage:
```
fpacifici, line 10: This seems conflicting with the implementation you provided here https://github.com/getsentry/arroyo/pull/57/files#diff-7ca52af34de9ab8fed8d220da046d4023236fcdbadc3a979718477f289d04713R11

Which one is the most up to date one ?

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
fpacifici: The idea was to make it as simple as possible for the user to have the dead letter queue in place. If something is broken just raise a specific exception. 
I agree that these parameters are quite ad-hoc. 
What about this:
```
class InvalidMessage(ABC):
   @abtrsactmethod
   def to_dict() -> Mapping


@dataclass
class InvalidKafkaMessage(InvalidMessage):
   consumer_group
   offset 
   partition
   timestamp
   payload

   def to_dict() -> Mapping:
       impl

class InvalidRawMessage(InvalidMessage): 
   payload

  def to_dict() -> .....
```

So we have typed implementations and no ad-hoc optional parameters
The exception would contain a list of `InvalidMessage`

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -225,12 +231,40 @@ def parallel_transform_worker_initializer(
...
+    be passed back to the function to start processing
+    from there.
+
+    `output_batch` contains all successfully processed
+    messages.
+
+    `bad_messages` contains `InvalidMessage` objects
+    caught during processing. These messages should be
+    passed to a DLQ by raising the `InvalidMessages`
+    exception.
```
fpacifici, line 253: you do not need these comments.
Call them `valid_messages` and `invalid_messages`. So that names and types say everything without the need of a redundant explanation.
At most you can add a few words to the description of the class explaining that it carries both valid messages and invalid ones.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -384,17 +425,25 @@ def __check_for_results(self, timeout: Optional[float] = None) -> None:
...
 
         logger.debug("Completed %r, reclaiming blocks...", input_batch)
         self.__input_blocks.append(input_batch.block)
-        self.__output_blocks.append(output_batch.block)
+        self.__output_blocks.append(result.output_batch.block)
         self.__batches_in_progress.decrement()
 
         del self.__results[0]
+        if bad_messages:
+            raise InvalidMessages(bad_messages)
```
fpacifici, line 446: Watch out, this method is called by `join` as well. IS it ok if you raise `InvalidMessages` there ?

arroyo/processing/strategies/streaming/transform.py
fpacifici: Then it seems we are having the wrong interface here. 
This method is called with an input batch, an index and is supposed to return a batch of processed message and a list of errors. 
For consistency, if you provide the list of invalid messages to append new errors to, you should also provide a batch of successful message to fill in so that you have a full batch of successful message and a full list of bad messages the last time you call the method.

This would be very cumbersome and counter intuitive. 
I think you should return a `ParallelTransformResult` with the batch of successful messages (as today) and the list of bad messages from the part of the batch that you processed. So the output batch is coherent with the list of bad messages. 

Then you merge the list of bad messages in `_check_for_results`

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
rahul-kumar-saini: The other one is most up to date, I just cloned the bare minimum here to be "consistent" and can deal with merge conflicts.

Maybe I should pull out this class/concept into its own PR and have it merged first.

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: I think raising it here is fine since `poll` should raise. `join` can be modified to also collect these 'collected' exceptions and then raise one.

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: Refactored to bring the `invalid_messages` out into the results tuple instead of passing it back into the worker.

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -293,34 +318,35 @@ def __init__(
...
         self.__batch_builder: Optional[BatchBuilder[TPayload]] = None
 
-        self.__results: Deque[
+        self.__processes: Deque[
             Tuple[
                 MessageBatch[TPayload],
-                AsyncResult[Tuple[int, MessageBatch[TTransformed]]],
+                AsyncResult[ParallelTransformResult[TTransformed]],
+                MutableSequence[InvalidMessage],
             ]
```
rahul-kumar-saini, line 343: I tired to pull this main `Tuple` out into a dataclass like I did for the result, but for some reason I can't add generic type params to `AsyncResult` when it's part of a class.

Eg.
```python3
@dataclass
class ParallelProcess:
    async_result: AsyncResult[<some_type>]
```
raises this:
```python3
TypeError: 'type' object is not subscriptable
```

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
rahul-kumar-saini: I don't know if we need to add that much complexity since at the end the result will be the same in the DLQ topic. There isn't really a logical difference between the two dataclasses you suggest.

Eg.
```python3
InvalidRawMessage(payload=str({"consumer_group": "something", "offset": 2, ...}))
InvalidKafkaMessage(consumer_group="something", offset=2, ...)
```

Both of these will 'dictify' into the same thing pretty much, why not just have one class and let the user decide what they care about:
```python3
@dataclass
class InvalidMessage:
     payload: Serializable
```

rahul-kumar-saini: Eg. Metrics processors just have a `Mapping[str, Any]` as their message.

https://github.com/getsentry/snuba/blob/c568f271eaf5130bf790b2b4e7a429dcbaf6f259/snuba/datasets/metrics_bucket_processor.py#L33-L35

I'm sure that message is a `KafkaMessage` but they'd end up using `InvalidRawMessage` anyway for invalid messages.

rahul-kumar-saini: https://github.com/getsentry/arroyo/pull/64

arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
rahul-kumar-saini: https://github.com/getsentry/arroyo/pull/64

arroyo/processing/strategies/streaming/transform.py
rahul-kumar-saini: I have no idea, if I don't add generic types mypy complains about it. These types also help later on cause we know what type would result from the async job. Maybe there is another "type" that is meant to be used for an AsyncResult/ApplyResult?

evanh: Yeah, no need to change this. Just confusing how this is working.

lynnagara (approved):
Looks good to me

arroyo/processing/strategies/streaming/transform.py
```diff
@@ -339,13 +365,14 @@ def __submit_batch(self) -> None:
         assert self.__batch_builder is not None
         batch = self.__batch_builder.build()
         logger.debug("Submitting %r to %r...", batch, self.__pool)
-        self.__results.append(
+        self.__processes.append(
```
fpacifici, line 368: For the next time. Please consider keeping the cosmetic refactoring in a separate PR.
Reviewing would be easier if only the relevant changes were in the changeset.

tests/processing/strategies/test_streaming.py
```diff
@@ -373,38 +377,96 @@ def test_parallel_transform_worker_apply() -> None:
...
+        fail_bad_messages,
+        input_batch,
+        output_block,
+    )
+    # all 9 messages processed
+    assert result.next_index_to_process == 9
+    # 5 were bad, 4 were good
+    assert len(result.invalid_messages) == 5
+    assert len(result.valid_messages_transformed) == 4
+    smm.shutdown()
```
fpacifici, line 467: Please test the case where `start_index` is not 0. That is the interesting case to test and to cover.

tests/processing/strategies/test_streaming.py
fpacifici: This is concerning. Where is the `ChildProcessTerminated` get raised?
It may well be that, in the case where the parallel transform step raises an exception, something wrong happens in when sending the exception back from the subprocess to the main process.
We need to ensure that a subprocess does not terminate when raising the exception if you need to reuse it.

fpacifici (commented):
Please do not merge till we figure out the reason for the `ChildProcessTerminated`

tests/processing/strategies/test_streaming.py
rahul-kumar-saini: The issue was I was calling `.join()` before `.close()`. Inverting the calls to those methods fixed the issue, no exception occurs regardless of where I position the test.

fpacifici (approved):

PR #62, opened 2022-04-21
feat: Avoid unnecessarily recreating processing strategy

lynnagara:
If the assignment or revocation callback is called with no partitions
we don't need to close and recreate the processing strategy. This
happens frequently with incremental rebalancing.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -79,21 +79,24 @@ class StreamProcessor(Generic[TPayload]):
 
         def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
             logger.info("New partitions assigned: %r", partitions)
-            if self.__processing_strategy is not None:
-                _close_strategy()
-            _create_strategy()
+            if partitions:
+                if self.__processing_strategy is not None:
+                    _close_strategy()
+                _create_strategy()
 
         def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
             logger.info("Partitions revoked: %r", partitions)
-            _close_strategy()
 
-            # Recreate the strategy if the consumer still has other partitions
-            # assigned and is not closed or errored
-            try:
-                if self.__consumer.tell().keys() - set(partitions):
-                    _create_strategy()
-            except RuntimeError:
-                pass
+            if partitions:
+                _close_strategy()
+
+                # Recreate the strategy if the consumer still has other partitions
+                # assigned and is not closed or errored
+                try:
+                    if self.__consumer.tell().keys() - set(partitions):
+                        _create_strategy()
+                except RuntimeError:
+                    pass
 
         self.__consumer.subscribe(
             [topic], on_assign=on_partitions_assigned, on_revoke=on_partitions_revoked


arroyo/processing/processor.py
```diff
@@ -79,21 +79,24 @@ def _create_strategy() -> None:
 
         def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
```
nikhars, line 80: The signature of this method says that partitions cannot be optional. If what you mention is that partitions could be `None` then can you update the signatures of these functions

arroyo/processing/processor.py
```diff
@@ -79,21 +79,24 @@ def _create_strategy() -> None:
...
             logger.info("New partitions assigned: %r", partitions)
```
nikhars, line 81: Shouldn't this go inside the if partitions check as well?

arroyo/processing/processor.py
```diff
@@ -79,21 +79,24 @@ def _create_strategy() -> None:
...
-            if self.__processing_strategy is not None:
-                _close_strategy()
-            _create_strategy()
+            if partitions:
+                if self.__processing_strategy is not None:
+                    _close_strategy()
+                _create_strategy()
 
         def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
             logger.info("Partitions revoked: %r", partitions)
```
nikhars, line 88: Same comment as above

arroyo/processing/processor.py
rahul-kumar-saini: I don't think it would be `None`, just an empty dictionary.

arroyo/processing/processor.py
lynnagara: I think it's useful to know when the partition assigned callback gets called even if it's empty

arroyo/processing/processor.py
lynnagara: Yeah, it's an empty dictionary

arroyo/processing/processor.py
nikhars: Ok. I was under the impression that partitions could be optional. In that case, it would cause an exception.

arroyo/processing/processor.py
```diff
@@ -79,21 +79,24 @@ def _create_strategy() -> None:
 
         def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
             logger.info("New partitions assigned: %r", partitions)
-            if self.__processing_strategy is not None:
-                _close_strategy()
-            _create_strategy()
+            if partitions:
```
nikhars, line 82: A hypothetical scenario.

Let's say you get new partitions assigned while the consumer is shutting down. That means you have set shutdown_requested to True over here https://github.com/getsentry/arroyo/blob/main/arroyo/processing/processor.py#L223. Is there still a need to keep creating the strategy again? Or can we check for the flag and do nothing in those scenarios?

lynnagara: I spent some time deliberating about this. I decided not to add it because it shouldn't be necessary. During shutdown all partitions should have be revoked and thus the strategy will not be recreated anyway (since `self.__consumer.tell().keys() - set(partitions)` should be falsy.

lynnagara:
@nikhars Mind taking another look?

nikhars (approved):

PR #64, opened 2022-04-27
Revamped invalid message model

rahul-kumar-saini:
### Overview
- PR #61 and #57 both rely on a common implementation of what an Invalid Message looks like.
- Pulling out that definition to this PR so it can be merged separately and there is no confusion on what it looks like.
- This would be merged into those two PR's before they are merged to main so there are no conflicts.

### DIFF
--- a/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
+++ b/arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
@@ -1,5 +1,78 @@
+import base64
 from abc import ABC, abstractmethod
-from typing import Any, Sequence
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Mapping, Optional, Sequence, Union
+
+SerializedPayload = Union[str, bytes]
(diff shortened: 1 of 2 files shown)


arroyo/processing/strategies/dead_letter_queue/policies/abstract.py
```diff
@@ -1,5 +1,95 @@
...
+    """
+
+    payload: SerializedPayload
+    timestamp: datetime
+    topic: str
+    consumer_group: str
+    partition: int
+    offset: int
+    headers: Headers
+    key: Optional[bytes] = None
```
rahul-kumar-saini, line 72: Key is optional in the `KafkaPayload` class, leaving it as `Optional[bytes]` here as well.

rahul-kumar-saini:
Key is optional in the `KafkaPayload` class, leaving it as `Optional[bytes]` for `InvalidKafkaMessage` as well.

fpacifici (approved):

PR #67, opened 2022-05-03
fix(tests): Make tests pass on M1 Macs

mcannizz:
ARM-based Macs use a 16K page size, so any block of shared memory will be
at least this size, regardless of size actually requested. (Intel machines
use a 4K page size instead.)

Change tests that allocate shared memory to do so in multiples of 16K so
that these tests behave the same regardless of machine architecture, and
update the sizes used by test data accordingly.

### DIFF
--- a/tests/processing/strategies/test_streaming.py
+++ b/tests/processing/strategies/test_streaming.py
@@ -320,11 +320,11 @@ def test_message_batch() -> None:
     smm = SharedMemoryManager()
     smm.start()
 
-    block = smm.SharedMemory(4096)
-    assert block.size == 4096
+    block = smm.SharedMemory(16384)
+    assert block.size == 16384
 
     message = Message(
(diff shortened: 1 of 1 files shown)


tests/processing/strategies/test_streaming.py
```diff
@@ -320,11 +320,11 @@ def test_message_batch() -> None:
     smm = SharedMemoryManager()
     smm.start()
 
-    block = smm.SharedMemory(4096)
-    assert block.size == 4096
+    block = smm.SharedMemory(16384)
+    assert block.size == 16384
```
mcannizz, line 324: I also considered changing the test such that it works with any page size, something like:
```
assert block.size == max(resource.getpagesize(), 4096)
```
However this seemed to be both more work and harder to reason about so I opted to change the numbers instead.

lynnagara (approved):
Thanks

PR #68, opened 2022-05-17
fix(dlq): Policy closure added + Producer closed

rahul-kumar-saini:
### Overview
- The produce policy takes in a producer which is never closed, this leaves the consumer hanging on shutdown
    - Added `self.__producer.close()` to the produce policy join method, this fixes the hanging issue
- The produce policy takes in a producer which is not recreated upon consumer rebalancing
    - Introduced the policy closure to the Consumer Strategy Factory in order to recreate the policy entirely every time the create() method is called (eg when rebalancing)
- Introduced `close()` and `terminate()` methods to Policies to be used when DLQ is closed/terminated

### DIFF
--- a/arroyo/processing/strategies/dead_letter_queue/policies/produce.py
+++ b/arroyo/processing/strategies/dead_letter_queue/policies/produce.py
@@ -24,6 +24,7 @@ class ProduceInvalidMessagePolicy(DeadLetterQueuePolicy):
     def __init__(
         self, producer: Producer[KafkaPayload], dead_letter_topic: Topic
     ) -> None:
+        self.__closed = False
         self.__metrics = get_metrics()
         self.__dead_letter_topic = dead_letter_topic
         self.__producer = producer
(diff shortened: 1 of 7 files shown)


arroyo/processing/strategies/dead_letter_queue/policies/produce.py
```diff
@@ -61,9 +64,17 @@ def _produce(self, payload: KafkaPayload) -> None:
         )
 
     def join(self, timeout: Optional[float] = None) -> None:
+        self.__producer.close()
```
fpacifici, line 67: Please Close this after you have waited on all futures.
Now the producer may still be able to send you the callbacks after it was closed, but this is an implementation detail that may change. 
It is generally a safer option to close a resource when you are done using it.

fpacifici (changes requested):
Please see the comment inline

fpacifici (approved):

PR #73, opened 2022-05-24
feat: Prevent passing invalid next_offset

lynnagara:
The next_offset is used to seek and commit messages. It must always
be equal to the offset of the last processed message + 1. We shouldn't
allow any other value to be passed here.

This is a breaking change.


arroyo/backends/local/storages/file.py
```diff
@@ -27,6 +28,15 @@
 from arroyo.utils.codecs import Codec
 
 
+@dataclass(unsafe_hash=True)
```
evanh, line 31: I'm just curious, why `unsafe_hash` is used for these classes?

evanh (approved):

arroyo/backends/local/storages/file.py
lynnagara: I copied this over but I think it's needed otherwise the object won't be hashable.

PR #74, opened 2022-05-30
feat(docs) Start Sphinx docs

fpacifici:
Adds the boilerplate for the Sphinx docs for this library and adds a basic intro.


docs/source/intro.rst
```diff
@@ -0,0 +1,23 @@
+Arroyo is a library to build streaming application that consumes from
```
lynnagara, line 1:
```suggestion
Arroyo is a library to build streaming applications that consume from
```

Makefile
```diff
@@ -13,3 +13,7 @@ lint:
 
 tests:
 	pytest -vv
+
+arroyodocs:
```
lynnagara, line 17:
```suggestion
docs:
```
I think it's a bit redundant to prefix make command with the name of the library

lynnagara (approved):

PR #75, opened 2022-06-02
feat(processing): Pass through partitions using create_with_partitions

MeredithAnya:
**context**
There is a fair amount of context already in https://github.com/getsentry/snuba/pull/2762, but long story short - we are trying enable auto scaling of consumers. The PR just mentioned focuses on a specific consumers (subscription executors) but we'd like to be able to do this for as many consumers as makes sense

**partition aware strategies**
There is logic in many of the consumers that is derived by the number of replicas. When we scale up or down consumers, the number replicas will change which means the number of partitions an individual consumer handles changes. If the strategies themselves are aware of these changes (aka reassignment or revocation) then they can essentially tune themselves when autoscaling kicks in. 

**`create_with_partitions`**:
Users of this library can write their own `create_with_partitions` (instead of writing a `create` method) if the want to take advantage of the partitions being passed in. This will not affect existing implementations with `create` because the default behavior for `create_with_partitions` is just to call `create`.

_non arroyo considered solution_
An initial thought to have strategies be aware of their partitions is done in https://github.com/getsentry/snuba/pull/2745/ where I kind of attempt to see if there has been a changed in the number of overall partitions by periodic checking which partitions have seen a message recently-ish. I've listed reason I don't think that approach is great 

**Next Steps**:

definite:
- [ ] upgrade arroyo in snuba and update all snuba consumers to use `create_with_partitions`
- [ ] upgrade arroyo in sentry and update all sentry (arroyo based) consumers to use `create_with_partitions`

unsure:
* if there is a reason to keep `create` instead of making `create_with_partitions` as the new standard then there doesn't need to be anymore follow up steps. Otherwise we can introduce a breaking change that forces the `create_with_partitions` to always be implemented

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -98,7 +98,7 @@ class StreamProcessor(Generic[TPayload]):
                 # assigned and is not closed or errored
                 try:
                     current_partitions = self.__consumer.tell()
-                    if current_partitions.keys() - set(partitions):
+                    if len(current_partitions.keys() - set(partitions)):
                         active_partitions = {
                             partition: offset
                             for partition, offset in current_partitions.items()
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -97,10 +97,11 @@ class StreamProcessor(Generic[TPayload]):
                 # Recreate the strategy if the consumer still has other partitions
                 # assigned and is not closed or errored
                 try:
-                    if self.__consumer.tell().keys() - set(partitions):
+                    current_partitions = self.__consumer.tell()
+                    if current_partitions.keys() - set(partitions):
                         active_partitions = {
                             partition: offset
-                            for partition, offset in self.__consumer.tell().items()
+                            for partition, offset in current_partitions.items()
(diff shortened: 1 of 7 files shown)


arroyo/processing/processor.py
```diff
@@ -94,7 +96,12 @@ def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
                 # assigned and is not closed or errored
                 try:
                     if self.__consumer.tell().keys() - set(partitions):
-                        _create_strategy()
+                        active_partitions = {
+                            partition: offset
+                            for partition, offset in self.__consumer.tell().items()
```
lynnagara, line 101: This won't actually happen or cause issues today, but I think it's better to call `consumer.tell()` only once in this function and then use the result in both of these 2 places to avoid the risk of the value ever changing in between.

evanh (approved):
This seems a simple enough change. I would say that if we can deprecate the `create` function, we should do it. Leaving it around longer than it needs to be will just cause headaches down the road.

lynnagara (approved):

PR #78, opened 2022-06-09
feat: Support throttle option in commit function

lynnagara:
Currently most consumers that use this library need to manually throttle
commits, as it's important to commit offsets to Kafka only periodically
and not on, say, every message.

The goal of this change is to make it easier to write consumers that do
this. The commit callback now supports an additional `throttle_sec` parameter.
If passed, offsets will be staged but committing won't be done unless that
duration has elapsed since the last commit.


MeredithAnya:
It seems like in many places, we use both a max size and max time in terms of determining when to commit. If we rely just on the time, it could vary a lot of how many messages we are committing at a time - does it make sense to have some sort of count that we increment each time we stage (but not commit) so that if it's been a certain amount of messages we can commit even if we haven't reached the `throttle_sec`?

Also for debugging purposes I think it could be nice to know how many offsets were staged occurred during the `throttled_sec` time.

arroyo/processing/processor.py
```diff
@@ -102,15 +105,25 @@ def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
...
-            "Waited %0.4f seconds for offsets to be committed to %r.",
-            time.time() - start,
-            self.__consumer,
-        )
+
+        now = time.time()
+
+        if throttle_sec is not None and (
+            self.__last_committed is None or now - self.__last_committed > throttle_sec
+        ):
```
rahul-kumar-saini, line 119: I think if `throttle_sec` is None this function will never commit. The logic here should be something like:

```
if throttle_sec exists and the time threshold has not been hit:
    return

self.__consumer.commit_positions()
...

lynnagara: You're right, thanks

fpacifici:
I am not fully convinced making the commit time transparent to the user would make things that much easier. On the contrary it opens up a scenario for accidentally loosing consistency guarantees.

In most consumer I can think about, we commit right after having performed a write operation which may not be idempotent.  Committing right after a write operation ensures at least once delivery, and not waiting makes sure that the chances of loosing "at most once" are minimized. 

If we provide a throttling option I would expect applications to leverage that by invoking `commit` after processing each message. This increases the chances to loose the at most once guarantees as the consumer has more time to crash between the write and the commit. 

Overall I feel having to decide when to commit is forcing the application developer to think about the delivery guarantees, so not a bad thing as we cannot guess them.

Do you have in mind any specific use case we would address where this concern is not an issue ?

lynnagara:
@fpacifici Choosing the commit frequency is a trade off between the consistency guarantees and additional load that committing adds on Kafka/Zookeeper. The goal of this change is to make it easy for the user to make that trade off in their consumer without implementing a bunch of code themselves.

The primary use case I see for the throttle option is when messages are *not* being chunked into large batches like we do for ClickHouse writes but when each message is processed and commit() called immediately afterwards.

In this scenario, by default committing will happen way too frequently. But this is not right for building performant, high throughput consumers. Since Arroyo currently provides no way to make throttling easy, currently the user needs to implement the throttle function themselves manually in their code, which I don't think is ideal. Examples of consumers today where this is an issue and we'd want to use this function are the subscriptions scheduler, executor and replay consumer.

I disagree that just because an option is being provided (which not even the default) that it the commit time any less transparent by us just providing the option. The user would have to explicitly opt into this. But without it I feel we are making it too difficult to write a consumer that commits infrequently.

By providing the option (and documenting how to use it properly), I actually think we would be actually encouraging the user of this library to think a bit more about their commit frequency, more than we would be by not supporting it at all in the API or ever mentioning it.

lynnagara:
@MeredithAnya I would prefer to omit the max size, as I don't really see a strong use case for this and it just adds complexity to the API. In Snuba max size and max time are used but the main goal of that is to generate batches to write on storage. In those cases where batching is already being done, I think it can be fine to just commit after each batch and not use this option at all as they are already fairly infrequent.

lynnagara:
Superceded by https://github.com/getsentry/arroyo/pull/106

PR #79, opened 2022-06-10
ref: don't depend on mypy, upgrade and fix mypy

asottile-sentry:
this is currently preventing an upgrade of `mypy` in any repository which depends on it

arroyo doesn't depend on mypy at runtime, only as a test-time utility

I also upgrade mypy, pytest and fixed the issues


lynnagara (approved):
Thanks

PR #81, opened 2022-06-17
feat: Remove the synchronized consumer

lynnagara:
The synchronized consumer should not be part of Arroyo. It was previously
used only by Snuba's legacy subscriptions infrastructure. This is now
rebuilt using regular (non-synchronized) consumers for performance reasons.
Because of those performance reasons, this is not really a pattern we want to
encourage by providing it out of the box in this library.

`arroyo.utils.synchronized` is now `arroyo.utils.commit`. Importing from
`arroyo.utils.synchronized` is still temporarily supported and will be removed
in a future version of this library.


fpacifici (commented):
This is a breaking change.

Should we announce it first in the release notes ..... then remove it one hour later

lynnagara:
@fpacifici There's a couple of breaking changes coming soon (this and the follow up changes to https://github.com/getsentry/arroyo/pull/75). How about we combine them into a 1.0.0 release?

fpacifici (approved):
ok

PR #84, opened 2022-06-28
docs: Add additional documentation to aid in first-time setup

cmanallen:
I had some trouble running the example so I've added documentation to improve the experience for new users.

### DIFF
--- a/examples/transform_and_produce/README.md
+++ b/examples/transform_and_produce/README.md
@@ -59,3 +59,21 @@ Produce a message to the topic by typing in a message in the expected format and
 ```
 
 Now the message should appear in the `raw-topic` shell, followed by another message in the `hashed-topic` shell. The password field of the message in `hashed-topic` should be the SHA256 hash of the password field of the message we manually produced to `raw-topic`.
+
+### Troubleshooting
+
+**Installing and running the `sentry_kafka` docker container.**
+
+1. In the file `~/.sentry/sentry.conf.py` append (or update) the `SENTRY_EVENTSTREAM` variable and set it to `"sentry.eventstream.kafka.KafkaEventStream"`.
+2. Change directory to the `sentry` repository.
+3. Set the `SENTRY_USE_RELAY` variable to `True`.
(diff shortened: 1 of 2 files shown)


examples/transform_and_produce/README.md
```diff
@@ -59,3 +59,21 @@ Produce a message to the topic by typing in a message in the expected format and
...
+   - It can be found in `sentry/src/sentry/conf/server.py`.
+4. Run `sentry devservices up`.
+5. You should see the following in stdout (or similar).
+
+```
+> Pulling image '.../cp-kafka:6.2.0'
+> Starting EXISTING container 'sentry_kafka' (listening: ('127.0.0.1', 9092))
+```
+
+Congratulations! You can now run docker commands against the `sentry_kafka` container. If you are still having issues reach out on the `#discuss-eng-sns` Slack channel.
```
onewland, line 79: I think we should drop this last sentence -- presumably this library is/could be used by people outside Sentry.

onewland (commented):
This is kind of a tough call because we don't want `arroyo` to be tightly coupled to Sentry itself -- it's supposed to be a general-purpose streaming processing library. Though this example already assumes that you're running the sentry application, it doesn't mean we should continue down that path

Maybe we could change the overall example README to run a known Kafka docker image to remove the dependency on `sentry_kafka`?

lynnagara:
Agree with @onewland here. I don't think these details are relevant for this library. All that this library's tests and examples require is a running Kafka on 127.0.0.1:9092. I think we may want to include some configuration here to spin these up instead of depending on Sentry's.

cmanallen:
@lynnagara @onewland Because running the example requires service orchestration (and independence from the sentry project) I think a tool like docker-compose would be beneficial.  Would that change be welcome?

lynnagara:
@cmanallen That seems good to me. If we can share the same configuration with CI https://github.com/getsentry/arroyo/blob/main/.github/workflows/ci.yml#L59-L76 that would be great.

PR #85, opened 2022-06-29
feat: Add docker-compose orchestrated example script

cmanallen:
This pull adds a docker-compose convenience script for testing the Arroyo project.  It is decoupled from Sentry and contains documentation for running the example.

Notes:

* I wanted to simplify the process further by running the Arroyo script within the docker-compose.yml file.  However, it seems the Arroyo example causes Kafka to crash if it's not ready to receive connections.
* The script doesn't perfectly mirror our CI.  The ports are different and LISTENERs have been tweaked but it is mostly similar and broadly applies.

### DIFF
--- a/examples/transform_and_produce/Dockerfile
+++ b/examples/transform_and_produce/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:3.8.13
+
+RUN apt-get update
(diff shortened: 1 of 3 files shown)


rahul-kumar-saini (approved):
Nice, followed the instructions locally and it all seems to work!

PR #95, opened 2022-08-04
upgrade confluent-kafka to 1.9.0

asottile-sentry:


evanh:
Have you tested this with Snuba/Sentry?

lynnagara:
I believe this is just a maintenance release, but just to confirm - did you check the changelog for the Python library as well as the bundled librdkafka 1.9.2? Is there anything of note in there?

asottile-sentry:
I was able to run the snuba testsuite in sentry using this -- the changelog didn't look interesting.  snuba I couldn't get building on M1 -- which I'll probably be able to fix by prebuilding some wheels (which is what this upgrade is blocking somewhat-ironically!)

lynnagara (approved):

PR #97, opened 2022-08-17
fix: Use correct log level for fatal crashes [sns-1622]

untitaker:
This will cause the sentry SDK, if initialized, to pick up the crash and
send it to sentry.

signals/KeyboardInterrupt shouldn't be caught by `except Exception:`

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -132,7 +132,7 @@ class StreamProcessor(Generic[TPayload]):
 
             self._shutdown()
         except Exception as error:
-            logger.info("Caught %r, shutting down...", error)
+            logger.exception("Caught exception, shutting down...")
 
             if self.__processing_strategy is not None:
                 logger.debug("Terminating %r...", self.__processing_strategy)


arroyo/processing/processor.py
```diff
@@ -132,7 +132,7 @@ def run(self) -> None:
 
             self._shutdown()
         except Exception as error:
-            logger.info("Caught %r, shutting down...", error)
+            logger.exception("Caught exception, shutting down...")
```
fpacifici, line 135: We should not need to log this as an exception here as we are raising it again at line 145. What was the issue ?

untitaker: for some reason i don't fully understand, `sys.excepthook` is broken in the child process. i suspect that multiprocessing.Pool overwrites it to propagate the exception to the parent process.

for some _other_ reason, the exception that makes the parent process shut down is also not reported.

now i can choose where i want to instrument: do I want to fix instrumentation in the parent process, where the exception stacktrace is broken and local variables are missing (because multiprocessing is pickling the exception to make it cross process boundaries), or in the child process where i risk losing errors if the consumer crash-loops in the parent process?

our answer for celery integration is that we instrument in the child process because there we get better contextual data.

I think we might want to instrument both processes just to be safe, but it would inherently mean duplicated error reports, unless we remove the `raise` in the child process

untitaker:
@fpacifici I investigated three parts wrt arroyo multiprocessing:

* Why does a crash of the main process not report sentry errors. I still don't have an answer to that and I don't expect to find one. I definitely know that something is broken here but i cannot reproduce it in a clean dev environment.
* What can we do to get the *best* stacktrace and contextual data when `process_message` crashes in the metrics string indexer. I think this PR is the answer and it should be merged.
* Why don't metrics get reported in subprocesses. Fix is in https://github.com/getsentry/sentry/pull/37950

I would suggest we investigate double-sending of errors when it becomes a practical concern. In my experience of writing the celery integration, it is tricky to proactively suppress double-reporting of errors without risking losing some insight into what is going on.

nikhars (approved):
Looks fine to me. I agree that having the ability to catch the error has more value than duplicate exceptions being reported.

PR #98, opened 2022-09-12
feat: Remove deprecated synchronized module

lynnagara:
The commit codec should be imported from arroyo.commit instead.


lynnagara:
Done in https://github.com/getsentry/arroyo/pull/106

PR #101, opened 2022-09-16
fix(docs) Fix makefile

fpacifici:
We cannot call the Makefile target for docs `docs` as there is a directory named `docs`


volokluev (approved):

lynnagara (changes requested):
Can we add docs to PHONY target instead?

fpacifici:
That may work, let me see

fpacifici:
yeah, it works

lynnagara (approved):
Thanks

PR #104, opened 2022-09-28
fix(collector) Add parallel collect timeout

fpacifici:
The parallel collect thread timeout was hardcoded. 
A consumer started hitting it and crashlooping.
This sets it as a parameter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: install lint setup-git tests, docs
+.PHONY: install lint setup-git tests docs
 
 setup-git:
 	pip install pre-commit==2.13.0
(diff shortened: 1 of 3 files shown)


volokluev (approved):

lynnagara (approved):

bumarcell:
Is there a way to change this parameter in sentry/snuba configs? 
@fpacifici @volokluev @lynnagara 🙏

fpacifici:
Sorry for the late answer. Did not notice this being closed 
We did not build it yet, but it would happen here 
https://github.com/getsentry/snuba/blob/9b8fb78d1ccaf97e24c68b8ef0d24b67818f0251/snuba/consumers/consumer_builder.py#L231

PR #106, opened 2022-10-05
feat: Introduce commit policies

lynnagara:
This is an attempt to simplify throttling commits in consumers. Generally speaking it is not safe for high throughput consumers to commit every offset. Today in order to achieve this with Arroyo, every consumer needs to manually implement throttling by themselves (unless batching is being used). This is verbose and easy to get wrong. This change aims to make it easier to do the right thing.

With this change, the stream processor takes an optional commit policy argument. The commit policy specifies the minimum time or minimum messages between commits.

In order to avoid throttling during shutdown, the commit function also now supports a force flag to override the commit policy / throttling behavior. This should be passed during consumer shutdown sequence as we no longer want to throttle in that scenario and just want to commit the latest offsets.

This change is designed to be backward compatible. The default behavior if no commit policy is passed remains the same as it was previously (no throttling, every offset committed).

### DIFF
--- a/arroyo/backends/kafka/commit.py
+++ b/arroyo/backends/kafka/commit.py
@@ -1,0 +1,55 @@
+from datetime import datetime
+from typing import Optional
+
+from arroyo.backends.kafka import KafkaPayload
+from arroyo.commit import Commit
+from arroyo.types import Partition, Topic
+from arroyo.utils.codecs import Codec
+
+DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+
+class CommitCodec(Codec[KafkaPayload, Commit]):
+    def encode(self, value: Commit) -> KafkaPayload:
+        assert value.orig_message_ts is not None
+
+        return KafkaPayload(
+            f"{value.partition.topic.name}:{value.partition.index}:{value.group}".encode(
+                "utf-8"
+            ),
+            f"{value.offset}".encode("utf-8"),
+            [
+                (
+                    "orig_message_ts",
+                    datetime.strftime(value.orig_message_ts, DATETIME_FORMAT).encode(
+                        "utf-8"
+                    ),
+                )
+            ],
+        )
+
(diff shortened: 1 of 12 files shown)


docs/source/getstarted.rst
```diff
@@ -266,6 +266,25 @@ do it in the poll on the completed callback. When the consumer is stopped, or th
 revoked, we wait for all the missing callbacks to complete in the `join` method.
 
 
+Committing offsets
+==================
+Arroyo does not auto commit offsets. It is up to you to manually commit offsets when processing for that
+message is completed.
+
+The commit callback will be passed to processing strategy via `ProcessingStrategyFactory.create_with_partitions`.
```
volokluev, line 274: can we have a code example?

docs/source/getstarted.rst
```diff
@@ -266,6 +266,25 @@ do it in the poll on the completed callback. When the consumer is stopped, or th
...
+You should pass this to the strategy and have you strategy call this commit function once the rest of the message
```
evanh, line 275:
```suggestion
You should pass this to the strategy and have your strategy call this commit function once the rest of the message
```

arroyo/backends/kafka/commit.py
```diff
@@ -0,0 +1,55 @@
...
+                    datetime.strftime(value.orig_message_ts, DATETIME_FORMAT).encode(
+                        "utf-8"
+                    ),
+                )
+            ],
+        )
+
+    def decode(self, value: KafkaPayload) -> Commit:
+        key = value.key
+        if not isinstance(key, bytes):
```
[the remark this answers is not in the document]
lynnagara: I believe they are nullable. I should add that this is all being moved into the kafka backend without any code changes (this is to avoid circular dependencies and because this code is only relevant to the kafka specific backend anyway), and would prefer to avoid making functional changes here.

evanh (approved):

PR #109, opened 2022-10-07
feat: Add produce step to library

lynnagara:
One of the most common things we do in our consumers is produce to another topic then commit offsets. This adds a basic ProduceAndCommit strategy to the Arroyo library which can be incorporated into processing pipelines, so that this doesn't need to be reimplemented constantly by end users.

Also updated the example in this codebase to use this step.

Code is largely taken from https://github.com/getsentry/snuba/blob/f5f94e4ebe0ed2c4a1b993dcffea1b70b030105a/snuba/subscriptions/executor_consumer.py#L402-L499 (with modifications). A slightly modified version of this been running in Snuba prod for some time.


arroyo/processing/strategies/produce.py
```diff
@@ -0,0 +1,86 @@
...
+            message, future = self.__queue[0]
+
+            if not future.done():
+                break
+
+            self.__queue.popleft()
+
+            self.__commit(
+                {message.partition: Position(message.next_offset, message.timestamp)}
+            )
```
fpacifici, line 49: This will require a commit policy to be set on the stream processor in order not to commit every offset on high throughput consumers, which are most. 
Do you have a plan to ensure people who use this producer do set a proper commit policy ?

arroyo/processing/strategies/produce.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def submit(self, message: Message[TPayload]) -> None:
+        assert not self.__closed
+
+        if len(self.__queue) >= self.__max_buffer_size:
+            raise MessageRejected
```
fpacifici, line 55: Knowing that, as of now, `MessageRejected` are often not properly managed by a step that send messages to this one and that when the Processor receives these back pressure events, it goes into a tight cpu loop waiting, wouldn't you prefer preventing this occurrence from happening at all, and waiting on the queue in the `poll` method instead  so that this condition is never verified? This until we make the `MessageRejected` process better.

I'd be ok either way though.

arroyo/processing/strategies/produce.py
```diff
@@ -0,0 +1,86 @@
...
+        self.__closed = True
+
+    def terminate(self) -> None:
+        self.__closed = True
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        start = time.time()
+
+        # Commit all pending offsets
+        self.__commit({}, force=True)
```
fpacifici, line 71: I think you should commit after having waited for the callbacks. You can still receive errors from them.

arroyo/processing/strategies/produce.py
```diff
@@ -0,0 +1,86 @@
...
+        self.__closed = False
+
+    def poll(self) -> None:
+        while self.__queue:
+            message, future = self.__queue[0]
+
+            if not future.done():
+                break
+
+            self.__queue.popleft()
```
fpacifici, line 45: I think this is missing the error cases. 
It is not checking the result of the future in order to propagate an exception if the production failed.
If the producer failed, the future will contain an exception

fpacifici (commented):
Some comments in line

fpacifici (changes requested):
Specifically, see my comment on the placement of commit in the join method and on error management

arroyo/processing/strategies/produce.py
```diff
@@ -0,0 +1,86 @@
...
+from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
+from arroyo.types import Commit, Message, Position, Topic, TPayload
+
+logger = logging.getLogger(__name__)
+
+
+class ProduceAndCommit(ProcessingStrategy[TPayload]):
+    """
+    For each message received, produces a message to the topic being passed in.
+    Waits on the futures and only once completed, commits the offset of the message.
```
fpacifici, line 17: Please let's elaborate on this.
The API documentation will mostly be generated from docstring (or will be the docstring).
I would specifically cover the delivery guarantee (at least once), the error management, and what to do with `MessageRejected` exceptions.

arroyo/processing/strategies/produce.py
fpacifici: Though if you keep `MessageRejected` around, in this step you will have to fix the parallel transform step as it does not handle those exceptions:
https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/streaming/transform.py#L397-L403.
Without fixing that, this step would not be usable after a parallel transform.

arroyo/processing/strategies/produce.py
lynnagara: Updated

arroyo/processing/strategies/produce.py
lynnagara: No, there is no plan for this nor should there be. The commit policy design aims to make commit policy and strategies completely orthogonal concerns. All policies work with all strategies. Strategies should call commit as frequently as they can (i.e. on every message like this one does) and it's the sole job of the commit policy to decide when to stage vs commit.

If you are worried about people picking the wrong default, I would be open to changing the default policy to once per second instead of on every message (The only reason I did not do that previously was to make commit policies a no-op when first introduced)

arroyo/processing/strategies/produce.py
lynnagara: Added it

arroyo/processing/strategies/produce.py
lynnagara: Yes, there an existing TODO in the parallel transform but that is not particularly relevant to this strategy. `MessageRejected` is the only proper Arroyo way of doing back pressure, I don't think we should do a workaround because of the bug in the other strategy.

arroyo/processing/strategies/produce.py
lynnagara: The comment may be misleading, I will update it. This is only committing already staged and succesfully produced offsets.

arroyo/processing/strategies/produce.py
fpacifici: I think it is relevant to this strategy as whoever creates a consumer cannot use a parallel transform strategy before this one. 
Could you call that out in the docstring? IT will be easy to miss that case when building a consumer.

lynnagara: Ok. I added it in both the docstring of this class as well as parallel tranform

fpacifici (commented):
Please see my comment on error management. The rest is ok

arroyo/processing/strategies/produce.py
```diff
@@ -77,6 +77,7 @@ def poll(self) -> None:
                             offset=message.offset,
                             headers=message.payload.headers,
                             key=message.payload.key,
+                            reason=str(exc),
```
fpacifici, line 80: For reason I mean pass the original exception:

```
raise WhateverException from exc
```

Still I think this should not be an InvalidMessage

arroyo/processing/strategies/produce.py
lynnagara: We set a default to make the whole thing a no-op and not require rewriting consumers when this was first introduced. I am happy to make it required in a future release and not have a default.

arroyo/processing/strategies/produce.py
lynnagara: Aren't all exceptions from strategies supposed to be InvalidMessage? DLQ needs to work with every strategy and that will not be possible if not.

fpacifici (approved):

arroyo/processing/strategies/produce.py
lynnagara: Ok: Made the change here https://github.com/getsentry/arroyo/pull/116. Will update all the usages within our codebases before merging this.

PR #111, opened 2022-10-07
feat: Add RunTask strategy

lynnagara:
This is designed to be one of the easiest possible strategies to use. It calls a function provided by the user on every message receives in a thread. Once completed, offsets are committed.

TODO:
- [ ] tests

### DIFF
--- a/arroyo/processing/run_task.py
+++ b/arroyo/processing/run_task.py
@@ -1,0 +1,77 @@
+import logging
+import time
+from collections import deque
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import Callable, Deque, Optional, Tuple
+
+from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
+from arroyo.types import Commit, Message, Position, TPayload
+
+logger = logging.getLogger(__name__)
+
+
+class RunTask(ProcessingStrategy[TPayload]):
+    """
+    Runs a task in a thread. Commits offsets once futures are done.
+    """
+
+    def __init__(
+        self,
+        processing_function: Callable[[Message[TPayload]], None],
+        concurrency: int,
+        commit: Commit,
+    ) -> None:
+        self.__executor = ThreadPoolExecutor(max_workers=concurrency)
+        self.__function = processing_function
+        self.__futures: Deque[Tuple[Message[TPayload], Future[None]]] = deque()
+        self.__max_pending_futures = concurrency * 2
+        self.__commit = commit
+        self.__closed = False
+
+    def submit(self, message: Message[TPayload]) -> None:
+        assert not self.__closed
+        # The list of pending futures is too long, tell the stream processor to slow down
+        if len(self.__futures) > self.__max_pending_futures:
+            raise MessageRejected
+
+        self.__futures.append(
+            (message, self.__executor.submit(self.__function, message))
+        )
+
+    def poll(self) -> None:
+        # Remove completed futures in order
+        while self.__futures and self.__futures[0][1].done():
+            message, _ = self.__futures.popleft()
+
+            self.__commit(
+                {message.partition: Position(message.next_offset, message.timestamp)}
+            )
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        start = time.time()
+
+        self.__commit({}, force=True)
+
+        while self.__futures:
+            remaining = timeout - (time.time() - start) if timeout is not None else None
+            if remaining is not None and remaining <= 0:
+                logger.warning(f"Timed out with {len(self.__futures)} futures in queue")
+                break
+
+            message, future = self.__futures.popleft()
+
+            future.result(remaining)
+
+            self.__commit(
+                {message.partition: Position(message.offset, message.timestamp)},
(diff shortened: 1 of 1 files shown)


arroyo/processing/run_task.py
```diff
@@ -0,0 +1,77 @@
...
+            raise MessageRejected
+
+        self.__futures.append(
+            (message, self.__executor.submit(self.__function, message))
+        )
+
+    def poll(self) -> None:
+        # Remove completed futures in order
+        while self.__futures and self.__futures[0][1].done():
+            message, _ = self.__futures.popleft()
```
fpacifici, line 44: I think we should elaborate a bit more this scenario.
Here we just check for completed futures and commit. We do not cover the case where the task fails and do not support a return value.
I would say managing the error case should be mandatory.
About the return value, I think that will be a need for most consumers as we generally either process a message and produce on a topic or batch processed message and write somewhere. The case where we just fork a task in a separate thread does not seem very common to me.

Also keep in mind that running a task in a separate thread is also not very common. We would not do that for high throughput consumers, we would have to batch messages first otherwise we would impact negatively our throughput with scheduling activity.

arroyo/processing/run_task.py
```diff
@@ -0,0 +1,77 @@
...
+
+from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
+from arroyo.types import Commit, Message, Position, TPayload
+
+logger = logging.getLogger(__name__)
+
+
+class RunTask(ProcessingStrategy[TPayload]):
+    """
+    Runs a task in a thread. Commits offsets once futures are done.
```
fpacifici, line 15: Do we have a scenario for this kind of consumer?

I think our common scenarios are:
- transformer: take a message, transform it and produce it somewhere 
- map-reduce: take a message, transform it, accumulate the result in a batch, periodically write.
- batch-transform-unbatch: accumulate messages, run a processing on batches, unbatch the results and produce on a topic.

All of them do something with the results of the processing.

arroyo/processing/run_task.py
lynnagara: Added the exception handling and some more documentation.

I intentionally avoided supporting a giving the user back a return value or chaining another step after this one. It's not meant for those use cases. It's only for running a simple IO-bound task and immediately committing offsets afterwards.

There are use cases for this already. I needed it here https://github.com/getsentry/sentry/pull/39502 for the post process forwarder but unfortunately it was not ready yet. I also saw some use cases where the batch processing strategy was being used where batching was not needed and a simple RunTask would have been better.

Yes - there are some scenarios where it's not appropriate, but it's not supposed to be an allen key that can fit every situation.

arroyo/processing/strategies/produce.py
```diff
@@ -101,7 +98,7 @@ def join(self, timeout: Optional[float] = None) -> None:
 
             future.result(remaining)
 
-            offset = {message.partition: Position(message.offset, message.timestamp)}
```
fpacifici, line 104: Was this committing the right offset ?

arroyo/processing/strategies/run_task.py
```diff
@@ -0,0 +1,97 @@
...
+        while self.__queue:
+            message, future = self.__queue.popleft()
+
+            if not future.done():
+                break
+
+            exc = future.exception()
+
+            if exc is not None:
+                raise exc
```
fpacifici, line 66: nit: if you call the `result` method on a future and the future contains an exception the exception will be raised. 
No need to extract the exception and re-rise it manually. `result` does it for you.

fpacifici (approved):

arroyo/processing/strategies/run_task.py
```diff
@@ -0,0 +1,99 @@
...
+
+from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
+from arroyo.types import Commit, Message, TPayload
+
+logger = logging.getLogger(__name__)
+
+TOutput = TypeVar("TOutput")
+
+
+class RunTask(ProcessingStrategy[TPayload]):
```
nikhars, line 15: Nit: I feel the current naming is too generic and could cause problems. How about we named it `RunTaskInThreads` to make it more explicit in what it does.

My primary concern is that in the documentation it is mentioned that `This strategy can be used to run IO-bound tasks in parallel, then commit offsets.` Parallel could mean in threads or processes. The current implementation is using threads. That means people should ensure that the processing_function should avoid using objects in the processing_function which can be modified by different threads or protect it using locks. Might be worth adding these caveats in the docstring as well.

arroyo/processing/strategies/produce.py
lynnagara: No, i actually messed it up earlier. Trying to stamp that out with position_to_commit

arroyo/processing/strategies/run_task.py
lynnagara: Updated

arroyo/processing/strategies/run_task.py
lynnagara: Good point, updated

PR #112, opened 2022-10-07
feat: Simplify example

lynnagara:
Since Arroyo provides transform and (soon) produce strategies there should be no reason to actually implement strategies own your own for simple use cases like the one in the example. This rewrites the example to leverage the existing strategies. Note: The produce part will be added separately as part of https://github.com/getsentry/arroyo/pull/109.


fpacifici (approved):

examples/transform_and_produce/README.md
```diff
@@ -10,17 +10,17 @@ The process is fairly simple:
 
 - The consumer consumes messages from the `raw-topic`
 - This message is in the form `{"username": "<username>", "password": "<password>"}`
-- The message is submitted to the `HashPasswordStrategy` which hashes the password string
+- The message is submitted to the `TransofrmStrategy` which hashes the password string
```
untitaker, line 13:
```suggestion
- The message is submitted to the `TransformStrategy` which hashes the password string
```

untitaker (approved):

PR #114, opened 2022-10-11
ref: Move strategies out of streaming module

lynnagara:
The streaming module is kind of redundant and just makes imports more verbose - this whole library is about streaming anyway and the batching strategies will be getting deprecated. This structure added confusion for users about when to use batching vs streaming. All the strategies will now live directly under the strategies module and are generally safe to combine together.


evanh (approved):

arroyo/processing/strategies/__init__.py
```diff
@@ -3,5 +3,17 @@
     ProcessingStrategy,
     ProcessingStrategyFactory,
 )
+from arroyo.processing.strategies.collect import CollectStep, ParallelCollectStep
+from arroyo.processing.strategies.filter import FilterStep
+from arroyo.processing.strategies.transform import ParallelTransformStep, TransformStep
```
nikhars, line 8: I wonder if in the future we need to have a hierarchy under strategies or not. Ideally, these should be independently composable. So there shouldn't be any restrictions on how to compose them. Validations on composition could happen outside of this module

nikhars (approved):

arroyo/processing/strategies/__init__.py
lynnagara: We don't need one. Isn't that exactly what this PR does though? All of them are independently composable and exported in this one place here for easy usage.

PR #116, opened 2022-10-13
feat: Make commit policy a required argument

lynnagara:
StreamProcessor will now require commit policy to be passed as a required argument. Previously the default (IMMEDIATE) was applied, however this is not a good default as it may lead to way too much committing. The user is now forced to pick a policy.

Also pulls out the docs on committing offsets into their own section, as it is easier to find than buried inside getstarted.


evanh (approved):

docs/source/offsets.rst
```diff
@@ -0,0 +1,50 @@
...
+            self.__commit = commit
+
+        def poll(self) -> None:
+            pass
+
+        def submit(self, message: Message[KafkaPayload]) -> None:
+            # do something (synchronous) with the message
+            do_something()
+            self.__commit(
+                {message.partition: message.offset_to_commit}
```
fpacifici, line 39: I wonder if one day we will want to change the signature of `commit` to just take the message object so there will be no more confusion on which offset to commit.

fpacifici (approved):

docs/source/offsets.rst
lynnagara: Damn, i just realise this line is wrong

lynnagara: Fwiw, I think the Position thing where the offset + timestamp gets passed every time is pretty suboptimal. It was needed because the commit log consumer is modelled as a consumer when the produce happens during the committing phase. I think this is quite wrong though and it should be modelled as a strategy instead ideally.

PR #117, opened 2022-10-18
docs(architecture) Adds an architecture page.

fpacifici:
Adds a simple architecture description.


fpacifici:
> it appears that sphinx is configured with the mermaid plugin, might be good to use that for the diagrams (since it's just markdown then)

Will try to redraw it there later.

PR #119, opened 2022-10-18
feat(routing): Add routing producer strategy

nikhars:
Add a strategy which is capable of routing messages to multiple producers and topics. It delegates determining the route on which to send a message to a router which can inspect messages and provide a producer and topic on which the message needs to be sent. This would be useful in scenarios like slicing where a single metrics indexer needs the flexibility to send messages to different slices of snuba consumers.


fpacifici:
Could you please beef up a bit the docstrings (they are the primary api docs) and add this module to the docs  ?
https://github.com/getsentry/arroyo/blob/main/docs/source/strategies.rst

arroyo/processing/strategies/routing_producer.py
```diff
@@ -0,0 +1,155 @@
...
+                f"Max size reached: total of {self.__messages_processed_since_last_commit} messages after {duration:.{2}f} seconds"
+            )
+            return True
+        if now >= (self.__started + self.__commit_max_batch_time):
+            logger.info(
+                f"Max time reached: total of {self.__messages_processed_since_last_commit} messages after {duration:.{2}f} seconds"
+            )
+            return True
+
+        return False
```
lynnagara, line 98: you should not implement commit throttling by yourself anymore, commit policies should handle that outside of the strategy. In here you can just call commit on every message.

arroyo/processing/strategies/routing_producer.py
```diff
@@ -0,0 +1,155 @@
...
+            (message, producer.produce(destination=topic, payload=message.payload))
+        )
+
+    def terminate(self) -> None:
+        self.__closed = True
+
+    def close(self) -> None:
+        self.__closed = True
+        for producer in self.__all_producers:
+            producer.close()
```
lynnagara, line 140: Since the producer is created outside the strategy and passed in, I'm not sure it's the strategy's job to close it. It would be easy to accidentally try to produce again on a closed producer. I think I would prefer either:
- have the strategy also create the producer, so closing here it is always safe
- close the producer outside wherever it gets created

nikhars:
> Could you please beef up a bit the docstrings (they are the primary api docs) and add this module to the docs ? https://github.com/getsentry/arroyo/blob/main/docs/source/strategies.rst

I added more documentation. If you have ideas on what additional things to document let me know. I am open to doing that.

arroyo/processing/strategies/routing_producer.py
```diff
@@ -0,0 +1,156 @@
...
+                break
+
+            exc = future.exception()
+            if exc is not None:
+                raise exc
+
+            self.__queue.popleft()
+            self.__offsets_to_be_committed[message.partition] = Position(
+                message.next_offset, message.timestamp
+            )
```
lynnagara, line 109: Please just call `self.__commit_function({message.partition: message.position_to_commit})` here and remove `self.__offsets_to_be_committed` entirely

arroyo/processing/strategies/routing_producer.py
```diff
@@ -0,0 +1,156 @@
...
+
+            message, future = self.__queue.popleft()
+            future.result(remaining)
+            offset = {message.partition: message.position_to_commit}
+
+            logger.info("Committing offset: %r", offset)
+            self.__commit_function(offset, force=True)
+
+        remaining = timeout - (time.time() - start) if timeout is not None else None
+        self.__message_router.shutdown(remaining)
```
lynnagara, line 156: You are still shutting down the message router here though (which I assume generally will be the one closing the producers). I don't think we should do this at all as they are not being created by this strategy. I think that whoever creates them should be the one calling close/shutdown on them.

arroyo/processing/strategies/routing_producer.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    A MessageRoute is a tuple of (producer, topic) that a message should be
+    routed to.
+    """
+
+    producer: Producer[KafkaPayload]
+    topic: Topic
+
+
+class MessageRouter(ABC):
```
lynnagara, line 26: This class contains quite a lot of boilerplate for my taste. I don't think it's all needed. I would actually prefer if this was essentially just the `get_route_for_message` function, ie. we just pass a `Callalble[[Message[KafkaPayload]], MessageRoute]` to the strategy. What do you think?

arroyo/processing/strategies/routing_producer.py
nikhars: I would disagree. If you look at the shutdown sequence, there is no way for the creator of the KafkaProducer to perform a clean close. A shutdown would trigger a clean close() of the consumer here. If you look at how a strategy is closed https://github.com/getsentry/arroyo/blob/main/arroyo/processing/processor.py#L61-L78, it just calls the close() and join() methods of the strategies. The processor factory would create the KafkaProducers here https://github.com/getsentry/arroyo/blob/main/arroyo/processing/processor.py#L39. But once they are created the control is lost.

Lets take this example https://github.com/getsentry/arroyo/blob/main/examples/transform_and_produce/__init__.py#L27-L52. The producer for this gets called from https://github.com/getsentry/arroyo/blob/main/examples/transform_and_produce/script.py#L24-L29. And the last step is processor.run(). All signal handling is performed by the processor. Where would the close for this producer happen in your opinion?

arroyo/processing/strategies/routing_producer.py
nikhars: Originally I tried to build it the way you suggested. But the abstract base class keeps thing simpler to implement.
1. There are multiple interfaces which need implementation. See my comment above. If the router is holding onto some resources, it needs to release them for a clean shutdown.
2. If there is state which you would like to maintain in the router you can do that. IMO, callables are ideal for stateless behavior. But for things like round robin router etc, you might want to maintain state.

arroyo/processing/strategies/routing_producer.py
lynnagara: Yes, the example is missing the close. I think closing should be done in the signal handler e.g. https://github.com/getsentry/snuba/blob/master/snuba/cli/subscriptions_executor.py#L162-L166. Otherwise producer will be closed after a rebalance and stop working after that.

arroyo/processing/strategies/routing_producer.py
lynnagara: Yeah, this is tightly related to the other comment, I'd prefer if shutdown wasn't the responsibility of the strategy at all.

arroyo/processing/strategies/routing_producer.py
nikhars: Since this is a blocker, rather than going back and forth on this, I am going to build this in Sentry to make progress on the project. I also feel like seeing the implementation end to end would provide more context and help clarify things much more easily. Going to close this PR.

PR #121, opened 2022-10-20
docs: Remove comment about asyncio in architecture

lynnagara:


fpacifici (approved):
I see you are against asyncIO ?

lynnagara:
@fpacifici Haha, not necessarily. Feel free to open an issue if you wanna discuss/push for it? I don't think it adds anything here.

fpacifici:
I think asyncIO can be very interesting but it is not that simple because, with the StreamProcessor, we actually tweak the way scheduling is done. Need more investigation.

PR #122, opened 2022-10-20
docs: Remove invalid release number

lynnagara:


onewland (approved):
seems fine -- did this have some kind of negative effect or it's just not being used?

PR #124, opened 2022-10-27
feat: Add stream processor metrics

lynnagara:
The purpose of this change is to provide some common basic metrics for all Arroyo consumers to help better understand consumer performance and break down where time is being spent.

Specifically 3 timings are being provided:
- `arroyo.consumer.poll.time` - Time spent in poll. Higher if the librdkafka buffer is empty
- `arroyo.consumer.processing.time` - Time spent in strategy submit and poll methods together
- `arroyo.consumer.paused.time` - Time spent in backpressure where consumer is paused

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -261,8 +261,6 @@ class StreamProcessor(Generic[TPayload]):
                             logger.info(
                                 "Paused for longer than %d seconds", LOG_THRESHOLD_TIME
                             )
-
-                        if paused_duration is not None:
                             self.__metrics_buffer.add_pause_time(paused_duration)
 
                 else:
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import logging
 import time
-from typing import Generic, Mapping, Optional, Sequence
+from collections import defaultdict
+from enum import Enum
+from typing import Generic, Mapping, MutableMapping, Optional, Sequence
 
 from arroyo.backends.abstract import Consumer
 from arroyo.commit import CommitPolicy
@@ -24,33 +26,28 @@ class InvalidStateError(RuntimeError):
     pass
 
 
+class ConsumerTiming(Enum):
+    CONSUMER_POLL_TIME = "arroyo.consumer.poll.time"
+    CONSUMER_PROCESSING_TIME = "arroyo.consumer.processing.time"
+    CONSUMER_PAUSED_TIME = "arroyo.consumer.paused.time"
+
+
 class MetricsBuffer:
     def __init__(self) -> None:
         self.__metrics = get_metrics()
         self.__reset()
 
-    def add_poll_time(self, duration: float) -> None:
-        self.__consumer_poll_time += duration
-        self.__throttled_record()
-
-    def add_processing_time(self, duration: float) -> None:
-        self.__processing_time += duration
-        self.__throttled_record()
-
-    def add_pause_time(self, duration: float) -> None:
-        self.__paused_time += duration
+    def increment(self, metric: ConsumerTiming, duration: float) -> None:
+        self.__data[metric] += duration
         self.__throttled_record()
 
     def flush(self) -> None:
-        self.__metrics.timing("arroyo.consumer.poll.time", self.__consumer_poll_time)
-        self.__metrics.timing("arroyo.consumer.processing.time", self.__processing_time)
-        self.__metrics.timing("arroyo.consumer.paused.time", self.__paused_time)
+        for (metric, value) in self.__data.items():
+            self.__metrics.timing(metric.value, value)
         self.__reset()
 
     def __reset(self) -> None:
-        self.__consumer_poll_time = 0.0
-        self.__processing_time = 0.0
-        self.__paused_time = 0.0
+        self.__data: MutableMapping[ConsumerTiming, float] = defaultdict(float)
         self.__last_record_time = time.time()
 
     def __throttled_record(self) -> None:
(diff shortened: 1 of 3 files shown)


arroyo/processing/processor.py
```diff
@@ -17,13 +17,48 @@
 
 logger = logging.getLogger(__name__)
 
-LOG_THRESHOLD_TIME = 10  # In seconds
+LOG_THRESHOLD_TIME = 5  # In seconds
```
lynnagara, line 20: The comment below in the implementation says 5 seconds anyway so just bringing this in line with that.

arroyo/processing/processor.py
```diff
@@ -17,13 +17,48 @@
...
 
 
 class InvalidStateError(RuntimeError):
     pass
 
 
+class MetricsBuffer:
```
fpacifici, line 27: I think this can be useful in many other contexts. It should be the default way to record metrics in a variable throughput environment. You could make it generic and reusable. Not a blocker.

arroyo/processing/processor.py
```diff
@@ -17,13 +17,48 @@
...
+    def __init__(self) -> None:
+        self.__metrics = get_metrics()
+        self.__metrics_frequency_sec = 1.0
```
fpacifici, line 30: Was this intended to be hardcoded? It is used in one place only. If it was meant to be hardcoded probably you should have a constant. If it was meant to be configurable it should be made configurable.

arroyo/processing/processor.py
```diff
@@ -17,13 +17,48 @@
...
+        self.__throttled_record()
+
+    def add_pause_time(self, duration: float) -> None:
+        self.__paused_time += duration
+        self.__throttled_record()
+
+    def flush(self) -> None:
+        self.__metrics.timing("arroyo.consumer.poll.time", self.__consumer_poll_time)
+        self.__metrics.timing("arroyo.consumer.processing.time", self.__processing_time)
+        self.__metrics.timing("arroyo.consumer.paused.time", self.__paused_time)
```
lynnagara, line 48: Just wanted to point out the metric naming convention being introduced here. I decided to prefix with `arroyo` as I think it keep things much better organized than they are now and avoids potential naming conflicts.

arroyo/processing/processor.py
```diff
@@ -177,15 +212,23 @@ def _run_once(self) -> None:
             # Otherwise, we need to try fetch a new message from the consumer,
             # even if there is no active assignment and/or processing strategy.
             try:
+                start_poll = time.time()
                 self.__message = self.__consumer.poll(timeout=1.0)
+                self.__metrics_buffer.add_poll_time(time.time() - start_poll)
```
fpacifici, line 217: What about making it a context manager like we do in Sentry ? 
https://github.com/getsentry/sentry/blob/6099b7d1f450ff594e8e651896a1d574c23bf451/src/sentry/utils/snuba.py#L278-L284

It would save a statement only, this is true, but in this method I think we should get everything we can out of the way as it is already complicated enough. So the context manager could help a bit readability.

arroyo/processing/processor.py
```diff
@@ -218,24 +261,36 @@ def _run_once(self) -> None:
                             logger.info(
                                 "Paused for longer than %d seconds", LOG_THRESHOLD_TIME
                             )
+
+                        if paused_duration is not None:
+                            self.__metrics_buffer.add_pause_time(paused_duration)
```
fpacifici, line 266: This will be overcounting drastically.
```
paused_duration = current_time - self.__paused_timestamp
```
or
```
paused_duration = current_time - self.__last_log_timestamp
```


So, as long as the consumer stays paused, `self.__paused_timestamp` does not increase while `current_time` does, so every time you get to this line you will log the entire time you have been paused since the last time logger.info was called, not the time passed since the last iteration you called `add_pause_time`. 

`self.__last_log_timestamp` is meant to avoid logging a line hundreds of times per second as, when you are paused this is a tight loop. The processor keeps calling submit, and it keeps receiving the exception.

arroyo/processing/processor.py
```diff
@@ -218,24 +261,36 @@ def _run_once(self) -> None:
...
+                            if self.__last_log_timestamp is not None
+                            else self.__paused_timestamp
+                        )
+
+                        self.__metrics_buffer.add_pause_time(
+                            time.time() - last_recorded
                         )
+
+                        self.__paused_timestamp = None
+                        self.__last_log_timestamp = None
```
fpacifici, line 292: I am not sure you need this part. 
It only meaningful covers the resume call. Logging the interval between the current time and __last_log_timestamp or __paused_timestamp would overcounting for the same reason I mentioned above.

fpacifici (commented):
Please see my comments on how to log the paused time.

arroyo/processing/processor.py
```diff
@@ -218,24 +261,36 @@ def _run_once(self) -> None:
                             logger.info(
                                 "Paused for longer than %d seconds", LOG_THRESHOLD_TIME
                             )
```
fpacifici, line 263: Your metrics will be much more effective than this. What about just removing this. We added it to troubleshoot a very specific issue and we never used this log since.

arroyo/processing/processor.py
lynnagara: Oops, this was supposed to go in the if statement above. Thanks!

arroyo/processing/processor.py
lynnagara: Yes it was meant to be hardcoded. It's only being used here right now.

arroyo/processing/processor.py
lynnagara: It is needed. Otherwise anything under 5 seconds won't get counted

arroyo/processing/processor.py
lynnagara: Yes. But I don't want to make it generic until we actually need it to be. It's not part of the public API so will not be a big deal to update later if we decide to reuse.

arroyo/processing/processor.py
fpacifici: I don't think this is correct. 
The loop works like this:
- get a message from kafka 
- submit to strategy
- MessageRejected
- pause consumer
- keep trying to submit to strategy as long as it keeps raising the exception

There is no pause or sleep in that loop.
So you will keep submitting and receiving the exceptions thousands of times per second.
So the `except` block keeps getting executed thousands of times per second. All these variables `pause_duration`, `self.__paused_timestamp`, `self.__last_log_timestamp` were only there in order to log every 5 seconds how long we have been paused.

I think you can remove all that complexity and simply accumulate pause time every time you go through the loop. So you flush the metric every second.

lynnagara: This is the part of the code where the consumer get resumed. I think it's important that the last pause metric gets recorded here immediately after consumer.resume() is called.. I needed to build it this way before to make both logging and metrics work together. Since we've decided to drop the log, I think this has been simplified quite a bit now.

arroyo/processing/processor.py
fpacifici: I think I am ok with that, just bear in mind that the client provides a metric wrapper object to begin with that generally has a prefix already

arroyo/processing/processor.py
fpacifici: I think this is still relevant

arroyo/processing/processor.py
```diff
@@ -197,45 +238,30 @@ def _run_once(self) -> None:
...
-                            and paused_duration > LOG_THRESHOLD_TIME
-                        ):
-                            self.__last_log_timestamp = current_time
-                            logger.info(
-                                "Paused for longer than %d seconds", LOG_THRESHOLD_TIME
+                        if self.__paused_timestamp:
+                            self.__metrics_buffer.add_pause_time(
+                                current_time - self.__paused_timestamp
                             )
+                            self.__paused_timestamp = current_time
```
fpacifici, line 249: better

fpacifici (approved):
Could you please look into my comment regarding the context manager before merging ?

lynnagara:
@fpacifici I ended up refactoring a little bit after you reviewed. About the context manager, I am sort of reluctant to provide 2 interfaces for the same thing right now given the current scope of this class. If there is only 1, I would prefer the non-context manager interface first as it seems a bit more versatile to me.

PR #129, opened 2022-11-08
fix: Make Position pickleable

lynnagara:
All of these common types should be pickleable. Since we expect the user might be passing Position around and manually passing it to the commit function, let's make it pickleable.

This is the workaround described in https://github.com/python/cpython/issues/80605

### DIFF
--- a/arroyo/types.py
+++ b/arroyo/types.py
@@ -67,11 +67,22 @@ class Message(Generic[TPayload]):
         return Position(self.next_offset, self.timestamp)
 
 
-@dataclass(order=True, unsafe_hash=True)
+@dataclass(frozen=True)
 class Position:
     __slots__ = ["offset", "timestamp"]
     offset: int
     timestamp: datetime
+
+    def __getstate__(self) -> dict[str, int | datetime]:
+        return dict(
+            (slot, getattr(self, slot))
+            for slot in self.__slots__
+            if hasattr(self, slot)
+        )
+
+    def __setstate__(self, state: dict[str, int | datetime]) -> None:
+        for slot, value in state.items():
+            object.__setattr__(self, slot, value)
 
 
 class Commit(Protocol):
--- a/arroyo/types.py
+++ b/arroyo/types.py
@@ -67,7 +67,7 @@ class Message(Generic[TPayload]):
         return Position(self.next_offset, self.timestamp)
 
 
-@dataclass(frozen=True)
+@dataclass(order=True, unsafe_hash=True)
 class Position:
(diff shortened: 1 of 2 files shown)


arroyo/types.py
```diff
@@ -67,7 +67,7 @@ def position_to_commit(self) -> Position:
         return Position(self.next_offset, self.timestamp)
 
 
-@dataclass(frozen=True)
+@dataclass(order=True, unsafe_hash=True)
```
untitaker, line 70: I think `order` is not strictly necessary?

untitaker (approved):
so if I understood correctly we're unfreezing the class here so that pickle works. I wonder if it would be better to define your own `__getstate__`/`__setstate__` instead. This would allow you to keep the class immutable for most purposes but bypass it for pickling specifically.

do we need `order`? I don't think the class had ordering previously.

arroyo/types.py
lynnagara: Yeah. I put it in because it matches the rest of the types in this class, and I think that ordering on int/timestamp field can make sense.

lynnagara:
@untitaker I wanted to use `unsafe_hash` to make it consistent with the way the rest of the data classes - Topic, Message, Partition - are done here. Though the question remains - should we then be changing all of them to use `__set_state__` and `__get_state__`?

nikhars:
> @untitaker I wanted to use `unsafe_hash` to make it consistent with the way the rest of the data classes - Topic, Message, Partition - are done here. Though the question remains - should we then be changing all of them to use `__set_state__` and `__get_state__`?

What was the original thought process which led to the decision of making these data classes have `unsafe_hash`? I don't think all the types defined by arroyo should have the same dataclass parameters for consistency. Each type has its own purpose which should drive how the dataclass is defined. Example: `Topics` are just strings which can be put into a set if needed. `Messages` on the other hand have TPayload which can be anything and hence should have `unsafe_hash`.

untitaker:
I agree with Nikhar, especially as somebody new to the codebase idk the answer to this.

If the only purpose of those flags was to make the dataclasses pickleable and they are to be treated as immutable otherwise, I would personally prefer a solution using `__getstate__`. This can also be hidden inside a class decorator.

lynnagara:
I don't know what the original thought process here was, this code has been this way before my time (it was copied verbatim into arroyo from snuba). I updated to use __get_state__ and __set_state. I do find this workaround excessively verbose though and preferred the original way even if it's not frozen.

untitaker (approved):
ok thanks, I should've made it more explicit that I don't really _strongly_ object

Issue #125, opened 2022-10-29
Make Arroyo consistent in the usage of the Message object


fpacifici:
There are two competing requirements in how processing strategies are designed. These are emerging from real world usages of the API.
On one hand, there are designs where there is a 1:1 relationship between messages seen by a `ProcessingStrategy` and messages produced by Kafka. The payload may have been altered but the Message still contains a meaningful partition and offset. The snuba consumer works like this
On the other hand there are designs where new messages are created in the pipeline that do not have a direct correspondence to an individual Kafka message. An example is the batching step in the Metrics Indexer. It accumulates messages into a batch and send the batch as a message to the next step. In this case the Batch, which has to be represented as a message, does not have a corresponding meaningful Kafka partition and offset.

The interface we have today seems to be designed for the first use case and does not work for the second.
This has implications:
- We cannot have a clean representation of a pipeline that contains aggregation steps or steps that introduce new messages as these cannot be associated to an individual Kafka message. Or if they do, that association has no meaning. This is troublesome as batching steps tend to be very useful. 
- Following the first approach means we cannot modularize pipeline as a sequence of steps that much. Any step that performs batching needs to encapsulate all the following steps if it wants to stick to the ProcessingStrategy API correctly. This is because it cannot assign an offset to a Batch. 
- This is the reason the existing CollectStep is so cumbersome. 

I think we need to make a decision about how to modularize our pipelines and stick to it:
- If we want to stick to the current interface: "a processing strategy takes a Message that has a valid partition and offset", then we cannot compose pipelines by introducing steps that batch messages and forward a batch to the following step. This means we need to introduce a different composition mechanism that works at a different abstraction level so that we can still express pipelines as `transform`, `filter`, `batch`.
- If we want to properly break this constraint and allow steps to forward batches, then the `submit` interface needs to change to accommodate the case where we forward a message that does not correspond to an individual Kafka message.

In alternative we can give up on the idea to provide batching features out of the box and let the client implement their own solution. I would advise against this as it is such a common and useful features. A ton of use cases rely on batching in some way, adding friction for the customer that wants to use it would defy the purpose of the api which is to make the right thing easy.

lynnagara:
[...]

PR #134, opened 2022-11-10
feat: Split the Message interface to better support batching steps

lynnagara:
This is an attempt to address the problems described in https://github.com/getsentry/arroyo/issues/125. It would be a major breaking change of Arroyo's API.

Today the `Message` interface is tied to a specific broker message (with a single partition, offset and timestamp). This doesn't make a lot of sense for strategies where the Message does not actually represent a single message but a batch of messages that have been grouped together - or, really, any scenario where a message no longer maps cleanly to a single underlying message from the Kafka broker.

This solution involves splitting the `Message` interface apart into `Message` and `BrokerPayload` to better serve the two distinct purposes. Now `Message` only has a `committable` property which is a mapping of partitions and offsets. This committable allows the message to easily span partitions. Since committing offsets is generally always the last step of a processing pipeline after all processing steps are done, this remains as part of the Message interface - it needs to always be passed through to every strategy so we know which offsets to commit at the very end after we are done with all the processing steps.

However messages received from the broker actually do map to a singular partition, offset and timestamp. The separate `BrokerPayload` class is introduced to address this. The first strategy in the pipeline will get a `Message[BrokerPayload[TPayload]]`; but thereafter the user can just pass a `Message[TPayload]` so the user can choose to include or exclude the BrokerPayload fields in later steps as needed.

I think that the main downside of this approach is that we are now storing additional information in every message about the partition and offset even when we are only dealing with singular messages (the committable is separate to the message's partition and offset data).

### DIFF
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -43,7 +43,7 @@ from arroyo.errors import (
     OffsetOutOfRange,
     TransportError,
 )
-from arroyo.types import BrokerPayload, Partition, Position, Topic
+from arroyo.types import BrokerValue, Partition, Position, Topic
 from arroyo.utils.concurrent import execute
 from arroyo.utils.retries import NoRetryPolicy, RetryPolicy
 
@@ -396,7 +396,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
 
     def poll(
         self, timeout: Optional[float] = None
-    ) -> Optional[BrokerPayload[KafkaPayload]]:
+    ) -> Optional[BrokerValue[KafkaPayload]]:
         """
         Return the next message available to be consumed, if one is
         available. If no message is available, this method will block up to
@@ -452,7 +452,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
                 raise ConsumerError(str(error))
 
         headers: Optional[Headers] = message.headers()
-        consumer_payload = BrokerPayload(
+        consumer_payload = BrokerValue(
             KafkaPayload(
                 message.key(),
                 message.value(),
@@ -700,7 +700,7 @@ class KafkaProducer(Producer[KafkaPayload]):
 
     def __delivery_callback(
         self,
-        future: Future[BrokerPayload[KafkaPayload]],
+        future: Future[BrokerValue[KafkaPayload]],
         payload: KafkaPayload,
         error: KafkaError,
         message: ConfluentMessage,
@@ -714,7 +714,7 @@ class KafkaProducer(Producer[KafkaPayload]):
                     raise ValueError("timestamp not available")
 
                 future.set_result(
-                    BrokerPayload(
+                    BrokerValue(
                         payload,
                         Partition(Topic(message.topic()), message.partition()),
                         message.offset(),
@@ -728,7 +728,7 @@ class KafkaProducer(Producer[KafkaPayload]):
         self,
         destination: Union[Topic, Partition],
         payload: KafkaPayload,
-    ) -> Future[BrokerPayload[KafkaPayload]]:
+    ) -> Future[BrokerValue[KafkaPayload]]:
         if self.__shutdown_requested.is_set():
             raise RuntimeError("producer has been closed")
 
@@ -743,7 +743,7 @@ class KafkaProducer(Producer[KafkaPayload]):
         else:
             raise TypeError("invalid destination type")
 
-        future: Future[BrokerPayload[KafkaPayload]] = Future()
+        future: Future[BrokerValue[KafkaPayload]] = Future()
         future.set_running_or_notify_cancel()
         produce(
             value=payload.value,
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -452,7 +452,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
                 raise ConsumerError(str(error))
 
         headers: Optional[Headers] = message.headers()
-        consumer_payload = BrokerValue(
+        broker_value = BrokerValue(
             KafkaPayload(
                 message.key(),
                 message.value(),
@@ -462,9 +462,9 @@ class KafkaConsumer(Consumer[KafkaPayload]):
             message.offset(),
             datetime.utcfromtimestamp(message.timestamp()[1] / 1000.0),
         )
-        self.__offsets[consumer_payload.partition] = consumer_payload.next_offset
+        self.__offsets[broker_value.partition] = broker_value.next_offset
 
-        return consumer_payload
+        return broker_value
 
     def tell(self) -> Mapping[Partition, int]:
         """
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -43,7 +43,7 @@ from arroyo.errors import (
     OffsetOutOfRange,
     TransportError,
 )
-from arroyo.types import Message, Partition, Position, Topic
+from arroyo.types import BrokerPayload, Message, Partition, Position, Topic
 from arroyo.utils.concurrent import execute
 from arroyo.utils.retries import NoRetryPolicy, RetryPolicy
 
@@ -394,7 +394,9 @@ class KafkaConsumer(Consumer[KafkaPayload]):
 
         self.__consumer.unsubscribe()
 
-    def poll(self, timeout: Optional[float] = None) -> Optional[Message[KafkaPayload]]:
+    def poll(
+        self, timeout: Optional[float] = None
+    ) -> Optional[Message[BrokerPayload[KafkaPayload]]]:
         """
         Return the next message available to be consumed, if one is
         available. If no message is available, this method will block up to
@@ -450,18 +452,26 @@ class KafkaConsumer(Consumer[KafkaPayload]):
                 raise ConsumerError(str(error))
 
         headers: Optional[Headers] = message.headers()
-        result = Message(
+        consumer_payload = BrokerPayload(
             Partition(Topic(message.topic()), message.partition()),
             message.offset(),
+            datetime.utcfromtimestamp(message.timestamp()[1] / 1000.0),
             KafkaPayload(
                 message.key(),
                 message.value(),
                 headers if headers is not None else [],
             ),
-            datetime.utcfromtimestamp(message.timestamp()[1] / 1000.0),
+        )
+        result = Message(
+            consumer_payload,
+            {
+                consumer_payload.partition: Position(
+                    consumer_payload.offset, consumer_payload.timestamp
+                )
+            },
         )
 
-        self.__offsets[result.partition] = result.next_offset
+        self.__offsets[result.payload.partition] = result.payload.next_offset
 
         return result
 
@@ -699,7 +709,7 @@ class KafkaProducer(Producer[KafkaPayload]):
 
     def __delivery_callback(
         self,
-        future: Future[Message[KafkaPayload]],
+        future: Future[BrokerPayload[KafkaPayload]],
         payload: KafkaPayload,
         error: KafkaError,
         message: ConfluentMessage,
@@ -713,12 +723,12 @@ class KafkaProducer(Producer[KafkaPayload]):
                     raise ValueError("timestamp not available")
 
                 future.set_result(
-                    Message(
+                    BrokerPayload(
                         Partition(Topic(message.topic()), message.partition()),
                         message.offset(),
-                        payload,
                         datetime.utcfromtimestamp(timestamp_value / 1000.0),
-                    )
+                        payload,
+                    ),
                 )
             except Exception as error:
                 future.set_exception(error)
@@ -727,7 +737,7 @@ class KafkaProducer(Producer[KafkaPayload]):
         self,
         destination: Union[Topic, Partition],
         payload: KafkaPayload,
-    ) -> Future[Message[KafkaPayload]]:
+    ) -> Future[BrokerPayload[KafkaPayload]]:
         if self.__shutdown_requested.is_set():
             raise RuntimeError("producer has been closed")
 
@@ -742,7 +752,7 @@ class KafkaProducer(Producer[KafkaPayload]):
         else:
             raise TypeError("invalid destination type")
 
-        future: Future[Message[KafkaPayload]] = Future()
+        future: Future[BrokerPayload[KafkaPayload]] = Future()
         future.set_running_or_notify_cancel()
         produce(
             value=payload.value,
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -43,7 +43,7 @@ from arroyo.errors import (
     OffsetOutOfRange,
     TransportError,
 )
-from arroyo.types import BrokerPayload, Message, Partition, Position, Topic
+from arroyo.types import BrokerPayload, Partition, Position, Topic
 from arroyo.utils.concurrent import execute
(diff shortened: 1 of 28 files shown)


arroyo/types.py
```diff
@@ -32,31 +32,46 @@ class Message(Generic[TPayload]):
...
     """
 
-    __slots__ = ["partition", "offset", "payload", "timestamp"]
+    __slots__ = ["payload", "committable"]
 
-    partition: Partition
-    offset: int
     payload: TPayload
-    timestamp: datetime
+    committable: Mapping[Partition, Position]
```
fpacifici, line 38: What if `committable` was a property or a method. 
So you could have a smarter implementation than duplicating the data between `Message` and `BrokerPayload`.

```
def committable(self) -> Mapping[Partition, Position]:
    if isinstance(self.payload, BrokerPayload):
         return {self.payload.partition: self.payload.position}
    elif isinstance(self.payload, BatchPayload):
         return .... whatever it takes to get the proper mapping ....
    # or you could have the dict initialized when the instance is created so you do not need to create a dictionary every time `committable` is called.
```

A cleaner alternative.
- add `committable` as a method of the current `Message`.
- Have subclaasses of Message. The broker based one just returns a dictionary with one partition and one position. The other one is for batch messages.
This would break a bit the concept of payload we have today but I am not sure it would be a big change.

lynnagara: Updated to remove the duplication of offsets. Let me know what you think.

untitaker (commented):
new to the codebase, but this makes sense to me

MeredithAnya:
make sense to me although just want to check: there is not anything preventing the `BatchPayload` from being like a random blob of data, like it doesn't have to be a batch of broker payloads right?

lynnagara:
> make sense to me although just want to check: there is not anything preventing the `BatchPayload` from being like a random blob of data, like it doesn't have to be a batch of broker payloads right?

Yes, it is a generic, the payload can be anything. It simply allows for offsets across multiple partitions to be committed after processing of that message.

arroyo/processing/strategies/batching.py
```diff
@@ -143,15 +143,6 @@ def submit(self, message: Message[TPayload]) -> None:
...
         start = time.time()
 
-        self.__metrics.timing(
-            "receive_latency",
-            (start - message.timestamp.timestamp()) * 1000,
-            tags={
-                "topic": message.partition.topic.name,
-                "partition": str(message.partition.index),
-            },
-        )
```
fpacifici, line 153: What happened to this ?

arroyo/types.py
```diff
@@ -85,6 +89,66 @@ def __setstate__(self, state: dict[str, int | datetime]) -> None:
             object.__setattr__(self, slot, value)
 
 
+@dataclass(unsafe_hash=True)
+class BatchPayload(Payload[TPayload]):
```
fpacifici, line 93: I like the idea that a `Payload` carries a `committable` property that can be managed by the logic that performs the commit in an abstract way.

Though I am not fully getting the naming of the BatchPayload and BrokerPayload. I am not sure I understand the difference of the two concepts and I think we may want to be a bit more opinionated. 

The way this is written, `BrokerPayload` is semantically a subset of a `BatchPayload`. Which means I could just use `BatchPayload` everywhere even if the `__payload` attribute represented one message only. The only issue would be with the `teimstamp` which is not there on the batch payload. 
This is a bit confusing as it make me ask myself why does `BrokerPayload` even exists.

I think we should understand which scenario we want to allow. I can think of:

1. A single message from the broker 
2. A single message that derives from a message coming from the broker (same position but different payload)
3. A batch of messages from the broker 
4. A batch of derived messages 
5. An empty batch with some offset to commit ?
6. A message that is not tied in any way to a message coming from the broker (thus no offset and partition) ?

I think 1 and 2 can easily managed by the same data structure. That's what we do today.
3 and 4 can be managed by the same data structure.
6 Is an interesting question. Should we support anything like that ? I think we could be opinionated in saying no to that. All messages we pass around are related to something to be committed. Can you think about a legit use case for this? Potentially a new message we want to produce somewhere, but why would we pass it through strategies before? If we do not allow this scenario, we are precluding this use case. If we allow that we can allow for an empty committable dictionary.

5 Is another interesting case. I can think about a scenario where we want to propagate am offset to commit even if we filtered out all the messages but I am not sure the Payload abstraction is the right one or whether there should be a special case of Payload without payload for this scenario.

Moreover why not making the `BatchPayload` opinionated in holding a Sequence of `TPayload` ? 
If we do this, Is there any reason why we need a `BrokerPayload` and not having a `BatchPayload` with one element? Is it to make strategies simpler? Do you have a specific use case in mind.

arroyo/processing/strategies/batching.py
lynnagara: A message here is no longer necessarily a BrokerPayload with a singular partition, topic, offset.

arroyo/types.py
fpacifici: >  I think we do not want to enforce that it must be an Array of TPayload. It can equally represent a filtered message with no data but only offsets to commit.

Then why not make it crystal clear to the user by having three types of payload:
- OneMessage
- Batch (that is an array)
- EmptyMessage (that only contains offsets without payload). 

So we make it clear what should be used for what just by reading the name.

lynnagara: EmptyMessage is something that we may want in the future. But I think we cannot support easily support it today without fairly significant changes. I think the biggest challenge is that it means that every strategy effectively needs to know about EmptyMessage and handle it appropriately, including user defined ones. I think there are still too many open questions for this to be included now.

About the Batch as an array, I think there are other reasons too why we shouldn't do it:
- If you look at the batch implementation in https://github.com/getsentry/arroyo/pull/127/ the `MessageBatch` object is not an array and contains other useful metadata about the batch.
- From a typing perspective, I think it will also make life more complicated. All messages passed to any strategy conform to `Message[TPayload]` and the typechecker ensures that `TOutput` of one strategy the `Message[TPayload]` of the next strategy. If we force BatchPayload to be `Message[List[TPayload]]` then I don't see how our common message class will work anymore as the type would not line up
- In the end I think the user can already choose to make TPayload = List[T] and forcing the type into an array doesn't make their life any meaningfully better

fpacifici: > EmptyMessage is something that we may want in the future. But I think we cannot support easily support it today without fairly significant changes. I think the biggest challenge is that it means that every strategy effectively needs to know about EmptyMessage and handle it appropriately, including user defined ones

That's fair, though how do you envision the problem (passing offset to commit) being managed in the implementation after this PR? Do we expect strategies that perform commit on behalf of others to be able to interpret an empty batch as some offsets to commit without any content ?

lynnagara: At this time, if we want to have the ability to:
- filter stuff at the start of the pipeline (or anywhere)
- all strategies can still work with all others
- users can easily extend and build their own strategies but they must handle the empty message too
- actually commit at least some those filtered messages

... then the only solution I can think of involves every single strategy knowing about some new `EmptyMessage` type or value and handling it in a custom way. In many cases it's probably just forwarding to the next step.

In this scenario all we can do is try and make this as un-painful for users as possible but it does add at least a little complexity to every strategy.

Thoughts?

lynnagara:
Changed the base branch since this probably warrants a 2.0.0 release.

arroyo/types.py
fpacifici: > If we want to have the ability to:

I believe this should be considered a requirement. At least this part: "A strategy needs to be able to filter messages at any point (basic product requirements). Periodically filtered out offsets must be committed even if there is no fully processed message (otherwise you accumulate messages in the Kafka queue and you may end up hitting Kafka retention breaking the consumer). An empty message cannot be committed until all the previous messages have been fully processed (basic consistency).

I think your idea `EmptyMessage`  is a valid solution.

Now, do we even have any alternative?
- In order to make filtering hidden from the steps that follow the filtering one, the message would have to be discarded early in the pipeline
- In order for offset to be still committed in the right order, we cannot commit in the place where we discard 
- This means that whoever commits needs to have visibility of the filtered messages.

This implies that, either the filtered message go through the pipeline anyway in some way or the processor keeps track of all messages and commits on its own when a number of filtered message is accumulated without any real message being present. 
Option two would basically go against the idea of keeping the commit process visible to the strategy.

So I don't think we really have an alternative that does not entirely hide the commit process from the application developer.

lynnagara: No I don't think there's any alternatives, there might be a few options for how `EmptyMessage` gets implemented but the outcome is the same. Every processing step just needs to handle it and pass it along to the next. Anything else is wrong from a consistency perspective.

lynnagara: I added an issue for this here https://github.com/getsentry/arroyo/issues/143. Let's move the discussion there. I think that this new message interface might might potentially make this a little easier (we can send one filtered message across many partitions instead of one per partition), but the problem is mostly unrelated to this change.

arroyo/processing/strategies/collect.py
```diff
@@ -32,7 +21,7 @@ def __init__(
 
         self.__created = time.time()
         self.__length = 0
-        self.__offsets: MutableMapping[Partition, OffsetRange] = {}
+        self.__offsets: MutableMapping[Partition, Position] = {}
```
fpacifici, line 24: Any functional reason to replace `OffsetRange` with `Position` or is it just a simplification?

arroyo/processing/strategies/run_task.py
```diff
@@ -76,10 +73,21 @@ def poll(self) -> None:
...
+            payload: BasePayload[TResult]
+            if isinstance(message.data, BrokerPayload):
+                payload = BrokerPayload(
+                    result,
+                    message.data.partition,
+                    message.data.offset,
+                    message.data.timestamp,
+                )
+            else:
+                payload = Payload(result, message.committable)
```
fpacifici, line 85: This pattern risk to become concerning. 
I think we should try to avoid littering the code base with `isinstance` calls that run different logic for each type of payload. We will certainly have an incident sooner or later for forgetting one type when adding one payload type.

Options:
- there is a way with mypy to check exhaustiveness of an if block. https://hakibenita.com/python-mypy-exhaustive-checking. That is designed for enums, not sure we can do something with it for different implementations, but I would look into that. 
- have transformation functions that generate a `Message` from payload replacing some content (on the line of dataclass `replace` functions). At least this pattern will be in fewer places
- horrible (but likely better than this), add a `to_message` method in the payload class.

arroyo/processing/strategies/run_task.py
```diff
@@ -90,13 +98,24 @@ def join(self, timeout: Optional[float] = None) -> None:
...
```
fpacifici, line 114: See above

arroyo/processing/strategies/transform.py
```diff
@@ -65,14 +65,18 @@ def poll(self) -> None:
...
+                result,
+                message.data.partition,
+                message.data.offset,
+                message.data.timestamp,
             )
-        )
+            self.__next_step.submit(Message(broker_payload))
+        else:
+            payload = Payload(result, message.committable)
+            self.__next_step.submit(Message(payload))
```
fpacifici, line 79: Same issue I had above, this looks the same pattern. Transform the payload of `Payload` or `BrokerPayload` and package it into a message. We can put it inside the payload class so we do not replciate this pattern everywhere.

arroyo/types.py
```diff
@@ -85,6 +79,76 @@ def __setstate__(self, state: dict[str, int | datetime]) -> None:
...
+    def payload(self) -> TPayload:
+        pass
+
+    @property
+    def committable(self) -> Mapping[Partition, Position]:
+        pass
+
+
+@dataclass(unsafe_hash=True)
+class Payload(BasePayload[TPayload]):
```
fpacifici, line 93: These two classes will create a lot of instantiations on high volume consumers. 
I think profiling or a load test is mandatory or we could be cutting down the throughput of our consumers by an undetermined amount. 

I would suggest you run the indexer load test referencing this specific arroyo commit in the dependency file so you can test the change without having to release arroyo. the pypi requirements file can reference a github commit instead of a release.

arroyo/types.py
```diff
@@ -85,6 +79,76 @@ def __setstate__(self, state: dict[str, int | datetime]) -> None:
...
+        self.offset = offset
+        self.timestamp = timestamp
+
+    @property
+    def payload(self) -> TPayload:
+        return self.__payload
+
+    @property
+    def committable(self) -> Mapping[Partition, Position]:
+        return {self.partition: Position(self.next_offset, self.timestamp)}
```
fpacifici, line 145: Is there in the standard library any data structure that implements `Mapping` and that is cheaper to instantiate than a dictionary ? This happens thousands times per second on a consumer so even a minimal improvement here can have a meaningful impact.

fpacifici (commented):
I am ok on the overall design. Though I think we should do something more to avoid littering the codebase of this pattern:

```
if isinstance(x, BrokerPayload):
....
else:
....
```

This will cause an incident. Imagine what happens if the users of the library start doing the same in their strategies. Adding new types of payloads will be almost impossible then.

I added a few possible ways to deal with that inline.

arroyo/processing/strategies/collect.py
lynnagara: No. Lo was never accessed

arroyo/processing/strategies/run_task.py
lynnagara: Ok. Added the replace method to the `Payload` to support swapping out the data and preserving the rest

arroyo/processing/strategies/transform.py
lynnagara: Updated

arroyo/types.py
lynnagara: I ran a few tests locally. It is about 0.1 seconds slower per 100,000 messages on my machine. I think it's acceptable. Since we are only talking about the cost of an extra Python class, I am going to skip the full indexer load test.

arroyo/processing/strategies/batching.py
lynnagara: I doubt it, unless you want to keep it for Snuba's replacer. Put it back anyway.

arroyo/backends/kafka/consumer.py
```diff
@@ -450,20 +452,19 @@ def poll(self, timeout: Optional[float] = None) -> Optional[Message[KafkaPayload
                 raise ConsumerError(str(error))
 
         headers: Optional[Headers] = message.headers()
-        result = Message(
-            Partition(Topic(message.topic()), message.partition()),
-            message.offset(),
+        consumer_payload = BrokerPayload(
```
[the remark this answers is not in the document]
fpacifici: What about 
Message contains Payload contains Value ?

arroyo/types.py
```diff
@@ -85,6 +81,85 @@ def __setstate__(self, state: dict[str, int | datetime]) -> None:
             object.__setattr__(self, slot, value)
 
 
+class BasePayload(Generic[TPayload]):
+    @property
+    def payload(self) -> TPayload:
```
fpacifici, line 86: why not calling this value (or swap them, Value contains Payload)
I feel having a `Payload` class that contains a payload will be quite confusing.

arroyo/types.py
fpacifici: Please still run the load test once you have the new version of the api released. It only takes 20 minutes to run it.

arroyo/types.py
lynnagara: Value seems good to me. It's sufficiently far away from the Kafka concept of value where there's less room for confusion

PR #138, opened 2022-11-16
feat: Remove BatchProcessingStrategyFactory from Arroyo

lynnagara:
Removes the BatchProcessingStrategyFactory from Arroyo. Eventually the underlying BatchProcessingStrategy and AbstractBatchWorker should be removed too, but these are still being used so we need to allow more time for deprecation of those.

### DIFF
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -16,10 +16,7 @@ from typing import (
     TypeVar,
 )
 
-from arroyo.processing.strategies.abstract import (
-    ProcessingStrategy,
-    ProcessingStrategyFactory,
-)
+from arroyo.processing.strategies.abstract import ProcessingStrategy
 from arroyo.types import Message, Partition, Position, TPayload
 from arroyo.utils.metrics import get_metrics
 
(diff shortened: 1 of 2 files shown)


MeredithAnya:
Sorry remind me again why we are removing this completely? Were these the strategies that were just in place to make it easy to migrate to arroyo in the first place? Is there never a good use case to keep these around?

lynnagara:
> Sorry remind me again why we are removing this completely? Were these the strategies that were just in place to make it easy to migrate to arroyo in the first place? Is there never a good use case to keep these around?

Arroyo will not be providing any factories in the future since they are too tied to a specific usage or sequence of processing steps. We are focussing on providing highly generic processing steps that are composable so that they can be chained together in any combination to form complex processing strategies.

As for the batch processing strategy in general, the only two places we were using it never even needed batching, RunTaskInThreads would have probably been a better fit if it was around at the time. We are also currently rebuilding from scratch the batch processing abstractions in Arroyo (see https://github.com/getsentry/arroyo/pull/127 and https://github.com/getsentry/arroyo/pull/134) so this will be even less needed.

tests/processing/strategies/test_batching.py
```diff
@@ -1,17 +1,48 @@
...
+    BatchProcessingStrategy,
 )
-from arroyo.types import Message, Topic
+from arroyo.types import Message, Partition, Position, Topic
+
+TPayload = TypeVar("TPayload")
+TResult = TypeVar("TResult")
+
+
+class BatchProcessingStrategyFactory(ProcessingStrategyFactory[TPayload]):
```
evanh, line 24: Why is this still needed in the tests?

arroyo/processing/strategies/batching.py
```diff
@@ -233,32 +230,3 @@ def __flush(self) -> None:
...
-        self.__worker = worker
-        self.__max_batch_size = max_batch_size
-        self.__max_batch_time = max_batch_time
-
-    def create_with_partitions(
-        self,
-        commit: Callable[[Mapping[Partition, Position]], None],
-        partitions: Mapping[Partition, int],
-    ) -> ProcessingStrategy[TPayload]:
-        return BatchProcessingStrategy(
```
fpacifici, line 259: Are we keeping the `BatchProcessingStrategy` ?

lynnagara: Long term no and it is already marked deprecated. Since we are using it currently, i will leave that one around for a bit longer

tests/processing/strategies/test_batching.py
lynnagara: Mostly I was lazy to refactor the tests since it will be deleted soon. But I think i will make the change now as it's not really needed.

nikhars:
While I am ok with the general idea of removing the strategy factories in general from arroyo, there is value in having an example strategy which clients can look at when first they are getting started with arroyo. Are you planning on having an example in the docs for reference?

lynnagara:
> While I am ok with the general idea of removing the strategy factories in general from arroyo, there is value in having an example strategy which clients can look at when first they are getting started with arroyo. Are you planning on having an example in the docs for reference?

There is one in `arroyo/examples`. Probably more soon.

tests/processing/strategies/test_batching.py
lynnagara: Rewrote the tests so it can be removed

evanh (approved):

Issue #137, opened 2022-11-16
Move committing offsets into it's own strategy


lynnagara:
Since Arroyo supports building consumers with at least once guarantees committing offsets is always the last step in a chain of processing strategies.

Today committing offsets often happens within a strategy. For example: `ProduceAndCommit` produces messages then commits the offsets when messages are produced, and `RunTaskInThreads` runs a task then commits offsets when it is done.

This means that `ProduceAndCommit` and `RunTaskInThreads` can both only be used as the last step in a sequence of processing steps. Something like `Consume -> RunTaskInThreads -> ProduceAndCommit` would not be possible.

In order to make our strategies more flexible, we can make committing offsets a separate step. Every other strategy except the commit strategy takes a `next_step` which can be any other step step (if more processing needs to be done) or the new `CommitStep` if that was the last step and we are ready to commit offsets.

PR #140, opened 2022-11-16
feat: Introduce separate commit strategy

lynnagara:
Since Arroyo supports building consumers with at least once guarantees, committing is always the last step in a chain of processing strategies after all other actions have completed.

Many of the processing steps today include committing offsets such as `ProduceAndCommit`, `RunTaskInThreads` and `BatchProcessingStrategy`. These strategies today can only be used as a last step in a chain. On the other hand, others take a `next_step` argument  which means another step must come afterward and they cannot be the last step.

This change aims make all of the other strategies more flexible by moving committing into a separate strategy. This strategy should be the only one ever used as the last strategy in a sequence. Every other strategy will take a `next_step` and can thus be used at any stage in a sequence of strategies.

Closes https://github.com/getsentry/arroyo/issues/137.

### DIFF
--- a/arroyo/processing/strategies/commit.py
+++ b/arroyo/processing/strategies/commit.py
@@ -1,0 +1,33 @@
+from typing import Optional
+
+from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.types import Commit, Message, TPayload
+
+
+class CommitOffsets(ProcessingStrategy[TPayload]):
+    """
+    Just commits offsets.
+
+    This should always be used as the last step in a chain of processing
+    strategies. It commits offsets back to the broker after all prior
+    processing of that message is completed.
+    """
+
+    def __init__(self, commit: Commit) -> None:
+        self.__commit = commit
+
+    def poll(self) -> None:
+        pass
+
+    def submit(self, message: Message[TPayload]) -> None:
+        self.__commit({message.partition: message.position_to_commit})
+
+    def close(self) -> None:
+        pass
+
+    def terminate(self) -> None:
+        pass
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        # Commit all previously staged offsets
+        self.__commit({}, force=True)
--- a/arroyo/processing/strategies/produce.py
+++ b/arroyo/processing/strategies/produce.py
@@ -6,12 +6,13 @@ from typing import Deque, Optional, Tuple
 
 from arroyo.backends.abstract import Producer
 from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
+from arroyo.processing.strategies.commit import CommitOffsets
 from arroyo.types import Commit, Message, Partition, Position, Topic, TPayload
 
 logger = logging.getLogger(__name__)
 
 
-class ProduceAndCommit(ProcessingStrategy[TPayload]):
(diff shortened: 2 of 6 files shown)


evanh (commented):
Is there any validation that checks to make sure a commit step is always the last step? Or is that left to the user?

untitaker:
afaik each strategy can decide that for itself based on whether its ctor accepts a `next_step`, but i am not entirely sure

untitaker (approved):

arroyo/processing/strategies/commit.py
```diff
@@ -0,0 +1,33 @@
+from typing import Optional
+
+from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.types import Commit, Message, TPayload
+
+
+class CommitOffsets(ProcessingStrategy[TPayload]):
```
fpacifici, line 7: I wonder if we actually need this step at all. 
The strategy factory can provide any step with a commit function. This step does not add anything on top of it. 
I think hiding the commit into `ProductAndCommit` was a good idea as the value added it provided was knowing when to commit.
This step instead does not add anything and still requires the application code to know when to commit and decide. 

So I think the original commit function may be enough. We could maybe improve that so that it takes a message instead of a position.

lynnagara: A few things:
- IMO this step adds A LOT on top of a commit function. With this, every other strategy we provide with Arroyo is way more flexible as it no longer needs to be one of either a) a strategy that commits (the last one in the chain) or b) a strategy that needs another one after it. Today `Produce`, `RunTask` and `Collect` have the restriction that they must be the last step and nothing can come after it. With this anything can also be chained after those steps. Similarly with steps that take a `next_step`. Today it is impossible for them to be the last step in a chain but with this they can be. Every strategy we ship with the library is a lot more flexible than it was before. If a user provides a custom strategy they are of course still welcome to call commit directly in it, but this improves the reusability of all strategies that ship with this library.
- Remembering to call `commit({}, force=True)` in join is something everyone forgets to do. This abstracts that away so no one needs to do that anymore
- Hiding the commit function is not a good idea or something that Arroyo seeks to do. On the contrary, Arroyo makes a point of passing the commit function to the factory so that the user must decide when to explicitly call commit. The point of the library is to make this easy but never to hide it. The same applies to commit policies. We force the user to provide one (it is a required argument). It manages the throttling for you but the user is always in control and we make a point of not hiding it
- Personally I don't think taking a message instead of a position is really an improvement at all. Conceptually, committing only requires offsets so passing the whole message with other random unrelated data which can be of any shape on it feels odd to me. There are already scenarios today where the message's offset is not the one that is safe to commit and we actually pass other offsets in here (e.g. subscription scheduler) which would break if we start forcing the whole message to be passed. In the future when we have `message.committable` this may improve but it will break use cases we have today. In the end the only benefit seems like it's being able to call `commit(message)` instead of `commit(message.committable)` which IMO doesn't make much of a difference to end users since they are already to doing it the other way and after this PR that call will be hidden inside this strategy anyway. I don't think it's worth changing up the API for that.

fpacifici: Ok, makes sense

fpacifici (approved):

PR #147, opened 2022-11-23
feat: Add RunTask strategy

lynnagara:
This is the most basic strategy that Arroyo provides. The rationale behind this addition is to make the library even simpler and reduce the amount of boilerplate for simple use cases.

Specifically, I've noticed that many people are using the BatchProcessingStrategy and AbstractBatchWorker combination almost exclusively for non batching use cases, purely because it allowed for using Arroyo without requiring a custom processing strategy. This is supposed to provide an easy replacement in those scenarios that doesn't have the downside of
unnecessarily waiting for batches to get filled.

This implementation is almost identical to the previous transform step, however RunTask explicitly allows for side effects. Now TransformStep is simply an alias for RunTask which is kept for backward compatibility and for symmetry with the ParallelTransformStep.

### DIFF
--- a/arroyo/processing/strategies/run_task.py
+++ b/arroyo/processing/strategies/run_task.py
@@ -14,6 +14,42 @@ TPayload = TypeVar("TPayload")
 TResult = TypeVar("TResult")
 
 
+class RunTask(ProcessingStrategy[TPayload]):
+    """
+    Basic strategy to run a custom processing function on a message.
+    """
+
+    def __init__(
+        self,
+        function: Callable[[Message[TPayload]], TResult],
+        next_step: ProcessingStrategy[TResult],
+    ) -> None:
+        self.__function = function
+        self.__next_step = next_step
+
+    def submit(self, message: Message[TPayload]) -> None:
+        self.__next_step.submit(
+            Message(
+                message.partition,
+                message.offset,
+                self.__function(message),
+                message.timestamp,
+            )
+        )
+
+    def poll(self) -> None:
+        self.__next_step.poll()
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        self.__next_step.join(timeout)
+
+    def close(self) -> None:
+        self.__next_step.close()
+
+    def terminate(self) -> None:
+        self.__next_step.terminate()
+
+
 class RunTaskInThreads(ProcessingStrategy[TPayload]):
     """
     This strategy can be used to run IO-bound tasks in parallel.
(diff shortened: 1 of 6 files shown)


arroyo/processing/strategies/run_task.py
```diff
@@ -14,6 +14,42 @@
...
+        )
+
+    def poll(self) -> None:
+        self.__next_step.poll()
+
+    def join(self, timeout: Optional[float] = None) -> None:
+        self.__next_step.join(timeout)
+
+    def close(self) -> None:
+        self.__next_step.close()
```
lynnagara, line 47: This is the only functional change in this PR. We are now calling next_step.close() in the close method instead of the join method as transform did previously.

lynnagara:
@fpacifici @untitaker Mind reviewing?

arroyo/processing/strategies/run_task.py
```diff
@@ -13,6 +13,37 @@
...
 
 
+class RunTask(ProcessingStrategy[TPayload]):
+    """
+    Basic strategy to run a custom processing function on a message.
+    """
+
+    def __init__(
+        self,
+        function: Callable[[Message[TPayload]], TResult],
```
fpacifici, line 23: Food for thoughts, now that we have a class that represents the payload of a message, should we change the interface of the transform function to take a value in and spit a value out  for consistency ?

fpacifici (approved):
Should we have an example that shows how to wire a runtask to a commit to build the simplest consumer ? 

Also please do not forget to add the strategy to the docs so that it shows up 
https://github.com/getsentry/arroyo/blob/main/docs/source/strategies.rst

arroyo/processing/strategies/run_task.py
lynnagara: What's the rationale for returning a value instead? This way the transform function is the same as it was before... and it's probably easier for the `value.replace(x)` call to be done here by the transform step instead of every user needing to call it by themselves.

fpacifici: I was more interested in passing the value instead of the message. 
It is just about a more intuitive api. A transformation function takes a an object and returns an objects that has the same role (thus Value -> Value)

lynnagara: I see. I guess that's possible. I don't think the decision here is related to the changes to Value/Message that we introduced that much though, or that the calculus really changes now. Like, it could just as easily have been Message -> Message before but we decided not to?

lynnagara:
The module was already in the docs

arroyo/processing/strategies/run_task.py
fpacifici: >  it could just as easily have been Message -> Message before but we decided not to?

I cannot remember what the reason was.

PR #149, opened 2022-11-24
feat: Collect / ParallelCollect supports a next step

lynnagara:
Instead of committing offsets after collect is done, now a next step can be passed to collect. This enables the collect step to be more flexible as it no longer needs to be the last step in a chain of strategies.

The other immediate motivation for this is to support a refactor of Snuba's consumer with commit log, which needs to do something else (produce offsets) after the collect step (which writes to ClickHouse) but before committing.


evanh (approved):

arroyo/processing/strategies/collect.py
```diff
@@ -14,14 +15,12 @@ class Batch(Generic[TPayload]):
...
         self,
         step: ProcessingStep[TPayload],
-        commit_function: Callable[[Mapping[Partition, Position]], None],
     ) -> None:
         self.__step = step
-        self.__commit_function = commit_function
 
         self.__created = time.time()
         self.__length = 0
-        self.__offsets: MutableMapping[Partition, Position] = {}
```
untitaker, line 24: I don't think this variable is needed at all anymore, no?

untitaker (approved):

arroyo/processing/strategies/collect.py
lynnagara: Changed to non private. The strategy uses `batch.offsets` as the offsets that get committed.

arroyo/processing/strategies/collect.py
```diff
@@ -143,38 +141,41 @@ def terminate(self) -> None:
...
-    so that kafka offsets are written in order.
+    The threadpool has 1 worker since we want to ensure batches are processed sequentially
+    and passed to the next step in order.
     """
 
     def __init__(
         self,
         step_factory: Callable[[], ProcessingStep[TPayload]],
-        commit_function: Callable[[Mapping[Partition, Position]], None],
+        next_step: ProcessingStep[Batch[TPayload]],
```
fpacifici, line 163: what about allowing to add a next_step in alternative to the commit_function so that this change is backward compatible? 

We have some direct usages of a collect step in snuba 
https://github.com/getsentry/snuba/blob/5c8d59f0aaf59d96d92a037ffba2436d2683b030/snuba/consumers/strategy_factory.py#L105-L125

and sentry
https://github.com/getsentry/sentry/blob/f2f6d64b8c3f231dc5d487f5d0c98269cfae6140/src/sentry/sentry_metrics/consumers/last_seen_updater.py#L192

fpacifici (commented):
Strategies built using the ConsumerStrategyFactory/KafkaConsumerStrategyFactory will be backward compatible but those are not the only strategies relying on the `CollectStep`. You may want to make this change backward compatible for them.

arroyo/processing/strategies/collect.py
lynnagara: I would actually prefer to just do a breaking change here and get us onto the version we want faster (every strategy must take a next step), CommitOffsets is the only exception. The number of changes is manageable and the fix in each case is just a few characters. (just passing `(CollectOffsets(commit_function))` there instead of `commit_function`. Keeps the code simpler in the future.

fpacifici: ok, up to you

PR #151, opened 2022-11-30
feat: Add replace() method to message

lynnagara:
Saves a few characters when we want to replace a message's payload and avoids the user needing to know about the Arroyo "value" internals. Instead of writing `message = Message(orig_message.value.replace(new_value))` we can now write `message = orig_message.replace(new_value)`.


onewland:
probably a dumb question: why wouldn't value just be immutable? Do Message objects get reclaimed when arroyo finishes processing them or something?

lynnagara:
> probably a dumb question: why wouldn't value just be immutable? Do Message objects get reclaimed when arroyo finishes processing them or something?

The Message and the Value it contains are both immutable. The replace method on both returns a brand new instance of Message / Value

untitaker (approved):

Issue #142, opened 2022-11-17
Allow the Dead Letter Queue to record the original message 


fpacifici:
The arroyo DLQ can be applied to any Processing Strategy.
While this is a relatively simple design it does not solve the most common DLQ problem: accumulate invalid messages and replay them after a fix is applied.
This is because only the first transformation step in an Arroyo consumer has access to the original message generally. If an error is identified in a following step, only the partially processed message would be sent to the DLQ, thus making it impossible to replay the messages.

One option would be to leave this in the hands of the application developer and recommend that the original message would be passed through as well. I think this is gross. It would make it really cumbersome and discourage people from employing the DLQ which would be an anti pattern.

Alternative option:
make the DLQ more transparent than it is now. Each step wrapped by a DLQ would only provide offset and partition in the invalid message (or invalid messages) exception. The DLQ step would be wrapping the entire strategy and would keep track of the original message indexed by partition and offset. When the DLQ strategy sees an invalid message it would produce the original message on the DLQ topic. (we would need to figure out a good way to clean up this buffer when offsets are committed)

lynnagara:
Option 3: Rewind the consumer and re-consume the bad message, this time knowing it is bad ahead of time

How it would work:
- The DLQ is not a regular processing strategy but an optional parameter passed to the stream processor
- Any strategy can raise an InvalidMessage with a specific offset (the message body is not required though)
- As soon as the StreamProcessor sees the Invalid Message it terminates the processing strategy and rewinds the consumer back to the last committed offset.
- The consumer then processes all the messages normally but stops at what it now knows is the bad offset
- When it hits the bad offset, it passes it to the DLQ which produces and immediately commits that
- Then it recreates the strategy again and continues normally

From the stream processor's perspective it is a bit of a dance. Throughput will slow temporarily. But it hides the complexity entirely inside the Arroyo library so users don't have to care and the processing strategy interface is dead simple. There is no need to buffer messages or pass copies of them through all the steps.

What do you think?

lynnagara:
We did it

PR #154, opened 2022-12-06
feat: Add a  schema validation strategy

lynnagara:
Schema validation implemented as a consumer strategy. The KafkaMessageDecoder takes a Kafka payload and is responsible for decoding and validating the payload. Any codec can be passed to the decoder - the one included here currently is for json / jsonschemas.


arroyo/processing/strategies/decoder.py
```diff
@@ -0,0 +1,102 @@
...
+    Placeholder. May eventually be replaced by a DLQ specific exception
+    once that feature is finalized.
+    """
+
+    pass
+
+
+class Codec(ABC):
+    @abstractmethod
+    def decode(self, raw_data: bytes, validate: bool) -> object:
```
onewland, line 73: would it make sense to do something with generics here so that mypy can verify that your processing steps can handle the output of your Codec? not pushing for this if it doesn't fit the use cases, but object seems about as wide open as `Any`

lynnagara: Done, for the base class at least. JsonCodec still returns an object but this allows the user to provide other codecs with more precise types if they want

lynnagara:
Reimplemented using multiprocessing now

evanh (commented):
Just some questions for my own understanding:

It doesn't look like any of this code handles any validation errors, and relies on `RunTaskWithMultiProcessing` to surface the `InvalidMessages`. Is that going to remain the strategy?

Also, the `RunTaskWithMultiProcessing` has a note that it doesn't handle downstream error messages. Would this validation being inserted cause a problem in that regard?

onewland (approved):
This seems fine to me but it's a little hard to evaluate in a vacuum -- maybe worth adding this to an existing or introducing a new example consumer

lynnagara:
> Just some questions for my own understanding:
> 
> It doesn't look like any of this code handles any validation errors, and relies on `RunTaskWithMultiProcessing` to surface the `InvalidMessages`. Is that going to remain the strategy?
> 
> Also, the `RunTaskWithMultiProcessing` has a note that it doesn't handle downstream error messages. Would this validation being inserted cause a problem in that regard?

About error handling: the strategy just throws a `ValidationError` right now if it does not match the schema. It's up to the user to wrap this strategy in something that decides what to do with the error. Eventually we probably want to replace this exception with a standard DLQ one but I didn't do that yet since the DLQ is subject to change right now (https://github.com/getsentry/arroyo/issues/142)

About the `RunTaskWithMultiprocessing` not handling MessageRejected from subsequent steps. The validation itself won't be affected but it might be an issue if validation is done before some other `Produce` or `RunTaskInThreads` step that uses backpressure. Basically it applies to exceptions from steps chained afterwards. Though we really need to get this fixed soon so all of the steps can be used together in any combination without these caveats.

arroyo/processing/strategies/decoder.py
```diff
@@ -0,0 +1,128 @@
...
+from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.processing.strategies.run_task import RunTaskWithMultiprocessing
+from arroyo.types import Message
+
+T = TypeVar("T")
+
+
+class DecodedKafkaMessage(NamedTuple):
+    key: Optional[bytes]
+    decoded: object
```
lynnagara, line 20: This probably needs to become generic too now

lynnagara:
Checked in my attempt to make DecodedKafkaMessage generic too in case anyone can help me fix it 😀

arroyo/processing/strategies/decoder.py
```diff
@@ -0,0 +1,130 @@
...
+        validate: bool,
+        next_step: ProcessingStrategy[DecodedKafkaMessage[T]],
+        num_processes: int,
+        max_batch_size: int,
+        max_batch_time: float,
+        input_block_size: int,
+        output_block_size: int,
+        initializer: Optional[Callable[[], None]] = None,
+    ) -> None:
+        self.__task_runner = RunTaskWithMultiprocessing(
```
fpacifici, line 55: Why hardcoding multiprocessing here? 
Multiprocessing adds a lot of overhead for cheap tasks on consumers that have a relatively low throughput. Moreover it makes the load on the downstream step very bursty. So often it may be counter productive. What about making it optional ?

arroyo/processing/strategies/decoder.py
```diff
@@ -0,0 +1,130 @@
...
+        decoded = rapidjson.loads(raw_data)
+        if validate:
+            self.validate(decoded)
+        return decoded
+
+    def validate(self, data: object) -> None:
+        try:
+            self.__validate(data)
+        except Exception as exc:
+            raise ValidationError from exc
```
fpacifici, line 130: Other problem with the parallel consumer. Actually we should definitively not use it here.
If a single message is bad in the parallel processing the whole batch is thrown away. I doubt this is what this is meant to do. 

Taking a step back. What do you plan to do with the ValidationError ? Should you raise an InvalidMessage instead that can be captured by the DLQ? The parallel transform in that case would let the good messages go on.

fpacifici (changes requested):
I don't think this use case is suitable for multi processing. 
The multi process transform step processes messages in batches. If one message in one batch raises an exception the whole batch is thrown out. Which is not what you want for a message validator that should be able to isolate bad messages.

lynnagara:
@fpacifici Good point, I reverted the multiprocessing part for now so the strategy can immediately reject on submit when a message that does not match the schema is received. I do still want to reintroduce a separate `DecoderWithMultiprocessing` later, but I think we should finalise how the new DLQ should work before doing that since there are still some open questions there.

untitaker (commented):
I wonder if it would be more useful, while we don't have DLQ, to validate a sample of the messages passing through, and log errors to Sentry for those. This would limit the performance overhead and would allow us to iron out bugs in schema validation or the underlying schema. I think that when we deploy this on existing infra, the schema is likely to reject messages that actually "pass through" perfectly fine.

fpacifici:
>  while we don't have DLQ

Technically we have a working DLQ. The main downside of the DLQ we have, in general, is that it may not give you the original message. But in the snuba consumers this is not a problem as there is only one transform step.

arroyo/processing/strategies/decoder.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        raise NotImplementedError
+
+
+class JsonCodec(Codec[object]):
+    def __init__(self, json_schema: object) -> None:
+        self.__validate = fastjsonschema.compile(json_schema)
+
+    def decode(self, raw_data: bytes, validate: bool) -> object:
+        decoded = rapidjson.loads(raw_data)
```
fpacifici, line 105: When are we going to play with simdjson ? 
https://simdjson.org/

fpacifici (approved):

arroyo/processing/strategies/decoder.py
lynnagara: Let me give a shot soon and see how it compares

PR #155, opened 2022-12-06
fix(commit_policy): Calculate elapsed timerange correctly

untitaker:
The first argument to should_commit is a duration/timedelta, not an
absolute timestamp.

This might explain why we are seeing severe performance regression when
using commit policy.

Some tests are failing, but i wanted to open this PR early on to get
some validation that i'm not totally confused?


untitaker:
I guess the failing test locally was a fluke. I should add tests to this before merging.

PR #158, opened 2022-12-10
fix: RunTaskWithMultiprocessing handles MessageRejected from subsequent steps

lynnagara:
Also moves the parallel transform tests into the run task module, since ParallelTransform is now just an alias for RunTaskWithMultiprocessing.


tests/processing/strategies/test_run_task.py
```diff
@@ -72,3 +95,363 @@ def test_run_task_in_threads() -> None:
...
+    assert invalid_message.payload == str(messages[0].payload)
+    # The 4 good ones should not have been blocked
+    assert next_step.submit.call_count == 4
+
+
+def add_one(message: Message[int]) -> int:
+    return message.payload + 1
+
+
+def test_message_rejected() -> None:
```
lynnagara, line 419: This is the only new test in this file. The rest is copied over from `test_transform` since ParallelTransform is now just an alias for RunTaskWithMultiprocessing.

evanh (approved):

PR #162, opened 2022-12-13
ref(commit_policy): Improve performance of commit policy

untitaker:
in QE environment, looking at the metrics in https://github.com/getsentry/sentry/pull/42291 show that using commit policies as part of https://github.com/getsentry/sentry/pull/41664 show a slowdown from avg 1 microsecond to 10 microseconds on every `Processor.commit` call

this PR moves a bunch of expensive logic from staging into commit to make the common case of not-committing faster, and skips a bunch of logic related to counting offsets if message-based commit policy is not configured

the result is that we are now at 4 microseconds avg instead, with the commit pre-master merge where #165 wasn't applied yet. I cannot test with that change because sentry fails to import Position then

the benchmarks only run once by default. to enable them, pass either `--benchmark-enable` to pytest to run them alongside tests, or `--benchmark-enable --benchmark-only` to run just the (currently singular) benchmark. Alternatively, run `make benchmarks`

There is also a workflow to compare benchmark results, using `--benchmark-save=baseline` and `--benchmark-compare=baseline`, see the documentation for pytest-benchmark or `pytest --help`.

### DIFF
--- a/arroyo/backends/local/backend.py
+++ b/arroyo/backends/local/backend.py
@@ -317,16 +317,6 @@ class LocalConsumer(Consumer[TPayload]):
             if self.__closed:
                 raise RuntimeError("consumer is closed")
 
-            if positions.keys() - self.__offsets.keys():
-                raise ConsumerError("cannot stage offsets for unassigned partitions")
-
-            self.__validate_offsets(
-                {
-                    partition: position.offset
-                    for (partition, position) in positions.items()
-                }
-            )
-
             self.__staged_positions.update(positions)
 
     def commit_positions(self) -> Mapping[Partition, Position]:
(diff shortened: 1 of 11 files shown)


arroyo/backends/local/backend.py
```diff
@@ -314,16 +314,8 @@ def seek(self, offsets: Mapping[Partition, int]) -> None:
...
-
-            if offsets.keys() - self.__offsets.keys():
-                raise ConsumerError("cannot stage offsets for unassigned partitions")
-
-            self.__validate_offsets(
-                {partition: offset for (partition, offset) in offsets.items()}
-            )
-
+            # XXX: can we remove the locking? dictionary updates might be
+            # atomic
```
lynnagara, line 318: Probably, the Kafka backend doesn't do this locking - messages need to be processed and committed in order. This one is just for testing anyway.

PR #163, opened 2022-12-13
fix: Ensure Collect/ParallelCollect properly calls next_step methods

lynnagara:
CollectStep and ParallelCollectStep were only calling the next_step.submit() method. The poll, close, join and terminate methods were never being called at all.


evanh (approved):

evanh:
Shouldn't there be tests that catch this?

PR #166, opened 2022-12-20
feat(decoders): Json decoder can be used without schema

lynnagara:
We want to be able to use the same JsonCodec in scenarios where we have a schema as well as scenarios where a topic schema is not yet available. This change makes schema optional. Validation function always returns true if no schema is provided.

### DIFF
--- a/arroyo/processing/strategies/decoder.py
+++ b/arroyo/processing/strategies/decoder.py
@@ -98,8 +98,11 @@ class Codec(ABC, Generic[T]):
 
 
 class JsonCodec(Codec[object]):
-    def __init__(self, json_schema: object) -> None:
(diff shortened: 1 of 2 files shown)


arroyo/processing/strategies/decoder.py
```diff
@@ -98,8 +98,11 @@ def validate(self, data: T) -> None:
 
 
 class JsonCodec(Codec[object]):
-    def __init__(self, json_schema: object) -> None:
-        self.__validate = fastjsonschema.compile(json_schema)
+    def __init__(self, json_schema: Optional[object] = None) -> None:
```
lynnagara, line 101: I mean technically None is still an object. But I wanted to be very explicit that this is nullable.

onewland (approved):

PR #169, opened 2023-01-05
auto publish to internal pypi on release

asottile-sentry:


evanh (approved):
@asottile-sentry What does this mean for the release process of this library now? Does it change yet?

asottile-sentry:
> @asottile-sentry What does this mean for the release process of this library now? Does it change yet?

same `craft` as usual, but it'll get put into internal pypi automatically

PR #172, opened 2023-01-12
build: Make json, msgpack and avro optional dependencies

lynnagara:
These are only needed if the user is using this strategy and a particular encoding format, and are likely not needed by the majority of users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ setup-git:
 	pre-commit install --install-hooks
 
 install:
-	pip install -e .
+	pip install -e .[avro,json,msgpack]
 	pip install -r requirements-test.txt
 
 lint:
--- a/arroyo/processing/strategies/decoder/json.py
+++ b/arroyo/processing/strategies/decoder/json.py
@@ -1,6 +1,6 @@
 from typing import Optional
 
-import fastjsonschema
+import fastjsonschema  # pip install sentry-arroyo[json]
 import rapidjson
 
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -2,6 +2,10 @@
 
 from setuptools import find_packages, setup
 
+AVRO_REQUIREMENTS = ["avro>=1.11.1"]
+JSON_REQUIREMENTS = ["python-rapidjson>=1.8", "fastjsonschema>=2.16.2"]
```
asottile-sentry, line 6: I would put these inline -- I don't think the constants give any additional benefits here

arroyo/processing/strategies/decoder/json.py
```diff
@@ -0,0 +1,26 @@
+from typing import Optional
+
+import fastjsonschema
```
asottile-sentry, line 3: one thing I like to do is put a helper comment on the first import so ti shows in the stacktrace -- like `# pip install sentry-arroyo[json]`

asottile-sentry (approved):

asottile-sentry (approved):

evanh (approved):

Makefile
```diff
@@ -5,7 +5,7 @@ setup-git:
 	pre-commit install --install-hooks
 
 install:
-	pip install -e .
+	pip install -e .[avro,json,msgpack]
```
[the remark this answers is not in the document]
lynnagara: Installs the optional dependencies (needed for test suite)

PR #173, opened 2023-01-17
feat: Remove CollectStep and ParallelCollectStep

lynnagara:
After https://github.com/getsentry/snuba/pull/3602, Collect and ParallelCollect will no longer used in Sentry and can be removed.

These strategies have already been marked as deprecated for some time because they are unnecessarily complex and unintuitive to use. In particular adopting the processing step interface for the batch accumulator made no sense. Arroyo already provides improved ways of achieving batching via the newer BatchStep, UnbatchStep and Reduce strategies.


fpacifici:
The CollectStep is still used in several places bot in Sentry and Snuba though or did we already replace it with Batch and Reduce? 
If we did not replace it yet, removing the Collect step means we would have to switch to the new step across the board in one deployment. I think that would be very risky as we never tested the new Batch step in terms of performance. 
I would try to replace the Collect step in some isolated places before (like try it on one Snuba consumer - one dataset, not one instance of a consumer) or better run a QE test on it, we can use the indexer first. 

Since we do not have a good utilization metrics for our consumers (specifically the snuba ones) it would be fairly hard to gauge the impact of this new step as, if it introduces a small degradation, it would eat into our capacity without us knowing.

lynnagara:
Collect/ParallelCollect no longer used anywhere - https://github.com/getsentry/snuba/pull/3602 was the last place.

Issue #143, opened 2022-11-18
Support committing offsets of filtered messages


lynnagara:
Today the FilterStep in Arroyo works by calling a `test_function` on every message received and either forwarding the message to the next step if it passes or dropping it if it does not.

Since committing must always happen as the last step in any pipeline, this means that the offsets of messages that are dropped by the filter will never be committed.

Most of the time this is not really a problem since we only ever want to commit offsets periodically anyway. Dropping some doesn't matter.

However in some use cases, so many messages might get filtered out that there may be an extended period of time for which no messages will ever get through the filter and nothing will get committed. This is problematic because the backlog will grow and alerts connected to this can fire.

One potential solution to this involves sending a special "filtered message" through the pipeline so it can get committed at the end. All the steps in between would have to recognise and simply forward the filtered message without doing any other processing on it. Though this has the downside that all other steps may need to have additional complexity as they would now need to be aware of the filtered message and handle it.

PR #177, opened 2023-01-24
fix(filter-strategy): Options to commit filtered messages [sns-1896]

untitaker:
Fixes https://github.com/getsentry/arroyo/issues/143

Message now contains a "filtered" variant that needs to be handled by every strategy.

`BaseValue[TPayload]`, however, can not contain a filtered message unless `TPayload is FilteredPayload`. This makes it possible to type `Reduce` in a way where users' callbacks  do not have to check for the presence of filtered messages to pass the typecheck.

**Breaking change**: All RunTask strategies now take callbacks that work on BaseValue rather than message, because BaseValue gives the guarantee that there is no FilteredPayload while Message doesn't.

### DIFF
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import time
 from collections import deque
-from typing import Deque, MutableSequence, Optional
+from typing import Deque, MutableSequence, Optional, cast
 
 from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
 from arroyo.processing.strategies.reduce import Reduce
@@ -102,13 +102,13 @@ class UnbatchStep(ProcessingStrategy[ValuesBatch[TPayload]]):
         next_step: ProcessingStrategy[TPayload],
     ) -> None:
         self.__next_step = next_step
-        self.__batch_to_send: Deque[BaseValue[TPayload]] = deque()
+        self.__batch_to_send: Deque[Message[TPayload]] = deque()
         self.__closed = False
 
     def __flush(self) -> None:
         while self.__batch_to_send:
             msg = self.__batch_to_send[0]
-            self.__next_step.submit(Message(msg))
+            self.__next_step.submit(msg)
             self.__batch_to_send.popleft()
 
     def submit(self, message: Message[ValuesBatch[TPayload]]) -> None:
@@ -116,8 +116,10 @@ class UnbatchStep(ProcessingStrategy[ValuesBatch[TPayload]]):
         if self.__batch_to_send:
             raise MessageRejected
 
-        assert not isinstance(message.payload, FilteredPayload)
-        self.__batch_to_send.extend(message.payload)
+        if isinstance(message.payload, FilteredPayload):
+            self.__batch_to_send.append(cast(Message[TPayload], message))
+        else:
+            self.__batch_to_send.extend(map(Message, message.payload))
         try:
             self.__flush()
         except MessageRejected:
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -6,7 +6,7 @@ from typing import Deque, MutableSequence, Optional
 
 from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
 from arroyo.processing.strategies.reduce import Reduce
-from arroyo.types import BaseValue, Message, TPayload
+from arroyo.types import BaseValue, FilteredPayload, Message, TPayload
 
 ValuesBatch = MutableSequence[BaseValue[TPayload]]
 
@@ -116,6 +116,7 @@ class UnbatchStep(ProcessingStrategy[ValuesBatch[TPayload]]):
         if self.__batch_to_send:
             raise MessageRejected
 
+        assert not isinstance(message.payload, FilteredPayload)
         self.__batch_to_send.extend(message.payload)
         try:
             self.__flush()
--- a/arroyo/processing/strategies/decoder/base.py
+++ b/arroyo/processing/strategies/decoder/base.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Generic, Optional, Sequence, Tuple, TypeVar
+from typing import Generic, Optional, Sequence, Tuple, TypeVar, cast
 
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.processing.strategies.abstract import ProcessingStrategy
@@ -18,7 +18,7 @@ class DecodedKafkaMessage(Generic[T]):
     headers: Sequence[Tuple[str, bytes]]
 
 
-class KafkaMessageDecoder(ProcessingStrategy[KafkaPayload]):
+class KafkaMessageDecoder(ProcessingStrategy[KafkaPayload], Generic[T]):
     """
     Decode messages to be forwarded to the next step. Optional validation.
     This strategy accepts a KafkaPayload and only performs validation on the
@@ -40,7 +40,7 @@ class KafkaMessageDecoder(ProcessingStrategy[KafkaPayload]):
 
     def submit(self, message: Message[KafkaPayload]) -> None:
         if isinstance(message.payload, FilteredPayload):
-            self.__next_step.submit(message.mark_filtered())
+            self.__next_step.submit(cast(Message[DecodedKafkaMessage[T]], message))
         else:
             decoded_value = self.__codec.decode(
                 message.payload.value, validate=self.__validate
--- a/arroyo/processing/strategies/decoder/base.py
+++ b/arroyo/processing/strategies/decoder/base.py
@@ -6,7 +6,7 @@ from typing import Generic, Optional, Sequence, Tuple, TypeVar
 
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.processing.strategies.abstract import ProcessingStrategy
-from arroyo.types import Message
+from arroyo.types import FilteredPayload, Message
 
 T = TypeVar("T")
 
@@ -39,14 +39,17 @@ class KafkaMessageDecoder(ProcessingStrategy[KafkaPayload]):
         self.__next_step.poll()
 
     def submit(self, message: Message[KafkaPayload]) -> None:
-        decoded_value = self.__codec.decode(
-            message.payload.value, validate=self.__validate
-        )
+        if isinstance(message.payload, FilteredPayload):
+            self.__next_step.submit(message.mark_filtered(DecodedKafkaMessage[T]))
+        else:
+            decoded_value = self.__codec.decode(
+                message.payload.value, validate=self.__validate
+            )
 
-        decoded = DecodedKafkaMessage(
-            message.payload.key, decoded_value, message.payload.headers
-        )
-        self.__next_step.submit(Message(message.value.replace(decoded)))
+            decoded = DecodedKafkaMessage(
+                message.payload.key, decoded_value, message.payload.headers
+            )
+            self.__next_step.submit(Message(message.value.replace(decoded)))
 
     def close(self) -> None:
         self.__next_step.close()
--- a/arroyo/processing/strategies/decoder/base.py
+++ b/arroyo/processing/strategies/decoder/base.py
@@ -40,7 +40,7 @@ class KafkaMessageDecoder(ProcessingStrategy[KafkaPayload]):
 
     def submit(self, message: Message[KafkaPayload]) -> None:
         if isinstance(message.payload, FilteredPayload):
-            self.__next_step.submit(message.mark_filtered(DecodedKafkaMessage[T]))
+            self.__next_step.submit(message.mark_filtered())
         else:
             decoded_value = self.__codec.decode(
                 message.payload.value, validate=self.__validate
--- a/arroyo/processing/strategies/filter.py
+++ b/arroyo/processing/strategies/filter.py
@@ -74,5 +74,6 @@ class FilterStep(ProcessingStep[TPayload]):
         self.__next_step.terminate()
 
     def join(self, timeout: Optional[float] = None) -> None:
+        self.__flush_uncommitted_offsets(time.time())
         self.__next_step.close()
         self.__next_step.join(timeout)
--- a/arroyo/processing/strategies/filter.py
+++ b/arroyo/processing/strategies/filter.py
@@ -1,6 +1,7 @@
 import logging
-from typing import Callable, Optional
+from typing import Callable, Optional, Union, cast
 
+from arroyo.processing.strategies.abstract import FILTERED_PAYLOAD, FilteredPayload
 from arroyo.processing.strategies.abstract import ProcessingStrategy as ProcessingStep
 from arroyo.types import Message, TPayload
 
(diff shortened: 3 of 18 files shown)


arroyo/processing/strategies/filter.py
```diff
@@ -29,7 +33,22 @@ def submit(self, message: Message[TPayload]) -> None:
...
+            self.__next_step.submit(cast_message)
+            self.__consecutive_drop_count = 0
+        else:
+            self.__consecutive_drop_count += 1
+
+            if (
+                self.__consecutive_drop_limit is not None
+                and self.__consecutive_drop_count >= self.__consecutive_drop_limit
+            ):
+                cast_message = cast(
```
untitaker, line 46: `message.replace()` returns `Message[FilteredPayload]` which does not fit into a `Message[Union[FilteredPayload, TPayload]]` shaped hole

therefore `Message` is missing covariance

arroyo/processing/strategies/run_task.py
```diff
@@ -42,23 +44,32 @@
...
         self.__next_step = next_step
 
-    def submit(self, message: Message[TPayload]) -> None:
-        result = self.__function(message)
-        value = message.value.replace(result)
-        self.__next_step.submit(Message(value))
+    def submit(self, message: Message[Union[FilteredPayload, TPayload]]) -> None:
+        if isinstance(message.value.payload, FilteredPayload):
+            self.__next_step.submit(
+                cast(Message[Union[FilteredPayload, TResult]], message)
```
untitaker, line 65: `message` is of type `Message[Union[FilteredPayload, TPayload]]`, and we're in a branch where it should've become `Message[FilteredPayload]` due to the conditional. That type should fit into `Message[Union[FilteredPayload, TResult]]`, but TPayload is invariant.

arroyo/processing/strategies/run_task.py
```diff
@@ -42,23 +44,32 @@
...
+            )
+        else:
+            result = self.__function(cast(Message[TPayload], message))
+            new_message = message.replace(result)
+            self.__next_step.submit(
+                cast(Message[Union[FilteredPayload, TResult]], new_message)
```
untitaker, line 71: Here we are trying to cast `Message[TResult]` to `Message[Union[FilteredPayload, TResult]]`

lynnagara (commented):
I think the typing of the strategies might not be quite right. There is no warning to the strategy author to handle the FilteredPayload in any way. Even worse - you actually have to cast the FilteredPayload in most cases to something that seems wrong to make it work properly.

Would something like adjusting the submit function to accept a `Union[Message[FilteredValue],  Message[TPayload]]` instead be a possibility? So it is made explicit that the user needs to handle this scenario.

arroyo/processing/strategies/decoder/base.py
```diff
@@ -39,14 +39,17 @@ def poll(self) -> None:
...
-        decoded_value = self.__codec.decode(
-            message.payload.value, validate=self.__validate
-        )
-
-        decoded = DecodedKafkaMessage(
-            message.payload.key, decoded_value, message.payload.headers
-        )
-        self.__next_step.submit(Message(message.value.replace(decoded)))
+        if isinstance(message.payload, FilteredPayload):
+            self.__next_step.submit(cast(Message[DecodedKafkaMessage[T]], message))
```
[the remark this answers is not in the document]
untitaker: this cast is there to convert `message` from `Message[TPayload]` to `Message[DecodedKafkaMessage[T]]`. This is fine because the payload is actually `FilteredPayload`.

This cast used to be hidden within a method on `Message`.

arroyo/processing/strategies/run_task.py
```diff
@@ -49,16 +50,19 @@ class RunTask(ProcessingStrategy[TPayload], Generic[TPayload, TResult]):
...
         next_step: ProcessingStrategy[TResult],
     ) -> None:
         self.__function = function
         self.__next_step = next_step
 
     def submit(self, message: Message[TPayload]) -> None:
-        result = self.__function(message)
-        value = message.value.replace(result)
-        self.__next_step.submit(Message(value))
+        if isinstance(message.value.payload, FilteredPayload):
```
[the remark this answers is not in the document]
untitaker: So that's what I had at the very beginning, I changed all strategies' baseclasses to read `ProcessingStrategy[Union[...]]`, Later it was suggested that we attempt to hide the filtered-ness in message, so the type of `message.payload` for a `message: Message[T]` is `Union[T, FilteredPayload]`.

The compiler does warn you about that. If this case isn't handled, pyright/mypy will warn you about `message.value.payload`/`message.payload` potentially being `FilteredPayload`.

arroyo/processing/strategies/produce.py
```diff
@@ -53,10 +63,14 @@ def poll(self) -> None:
...
             committable, future = self.__queue[0]
 
-            if not future.done():
+            if future is not None and not future.done():
                 break
 
-            message = Message(Value(future.result().payload, committable))
+            message: Message[TPayload]
+            if future is None:
+                message = Message(Value(FILTERED_PAYLOAD, committable))
```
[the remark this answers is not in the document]
untitaker: I didn't find a way to create a future from scratch that resolves immediately. Even creating an `async def` doesn't seem to work since it returns a coroutine instead.

arroyo/processing/strategies/filter.py
```diff
@@ -28,8 +39,30 @@ def poll(self) -> None:
     def submit(self, message: Message[TPayload]) -> None:
         assert not self.__closed
 
+        now = time.time()
+
         if self.__test_function(message):
+            if self.__commit_policy_state is not None:
+                self.__flush_uncommitted_offsets(now)
             self.__next_step.submit(message)
```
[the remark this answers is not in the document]
untitaker: >Since there are enough messages actually passing the test for committing to still regularly happen

But this PR is about scenarios where this isn't the case, right? You don't have to pass commitpolicy into FilterStep. If you don't pass one, FilterStep works exactly like it does on master, and that version of it is already pretty much perfect for that scenario.

----

I see three paths forward:

1. we expect that all partitions regularly receive messages that pass the filter. However, in that case one could just not pass a commit policy into FilterStep and the code works exactly as in master, which already works fine under such conditions.
2. we make the flushing conditional based on the given commit policy (just like in the codepath on line 51). this will cause offsets to be committed out of order, i.e. offset 2 can be committed after offset 3. we can make further changes in the processor or elsewhere to sanitize those commits
3. we attach the to-be-flushed committable offsets to the message that we submit immediately after this one (on line 47). This requires us to change the concrete type of `message.value` which strategies could rely on (even though they probably shouldn't)

Ideally the filterstep would be able to exhibit optimal performance regardless of whether the majority of messages are filtered, I just couldn't think of a way to make that happen.

untitaker:
> There is no warning to the strategy author to handle the FilteredPayload in any way.

`message.value.payload` is `Union[FilteredPayload, TPayload]`. If you don't explicitly handle the FilteredPayload variant of the union, you won't be able to directly access `TPayload`. If you don't directly access `TPayload`, there is no need to explicitly handle `FilteredPayload`. For example the `Commit` strategy doesn't have to distinguish filtered messages.

But otherwise mypy should yell at you pretty loudly to handle that case explicitly. For example you can't pass a `BaseValue[TPayload]` to the user's callback in `RunTask` without thinking about it, since the actual type you're getting is `BaseValue[Union[FilteredPayload, TPayload]`.

Yes, there are ugly casts everywhere... This is indeed unfortunate. Part of the problem here is that `TPayload` is still invariant, another problem is that sometimes we want to transform `Message[TPayload]` into `Message[TResult]`, and that can and should be a noop if the `Message` contains a `FilteredPayload`. We can hide some of it behind a function like this:

```
class Message(Generic[TPayload]):
    def assert_filtered(self) -> Message[TResult]:
        assert isinstance(self.payload, FilteredPayload)
        return cast(Message[TResult], self)
```

arroyo/processing/strategies/produce.py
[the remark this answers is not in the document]
lynnagara: can you do something like:
````
fut = asyncio.Future()
fut.set_result(FILTERED_PAYLOAD)
````

File: arroyo/processing/strategies/filter.py [hunk 8]
```diff
@@ -28,8 +39,30 @@ def poll(self) -> None:
...
+        elif self.__commit_policy_state is not None:
+            self.__uncommitted_offsets.update(message.committable)
+
+            if self.__commit_policy_state.should_commit(now, message.committable):
```
lynnagara, line 51: Could we move the `should_commit` check into the `__flush_uncommitted_offsets` method. That way, even in the other code branch (test function passes and message is valid) we still keep the same throttling behavior

arroyo/processing/strategies/run_task.py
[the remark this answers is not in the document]
lynnagara: Makes sense, I think i missed this earlier.

arroyo/processing/strategies/batching.py
```diff
@@ -116,6 +116,7 @@ def submit(self, message: Message[ValuesBatch[TPayload]]) -> None:
         if self.__batch_to_send:
             raise MessageRejected
 
+        assert not isinstance(message.payload, FilteredPayload)
```
fpacifici, line 119: Why asserting this is not a filtered payload? Does the unbatch job cares if it receives a filtered message? 
It seems to me it should just ignore them.

arroyo/types.py
```diff
@@ -39,13 +46,17 @@ class Message(Generic[TPayload]):
...
-    value: BaseValue[TPayload]
+    value: BaseValue[Union[FilteredPayload, TPayload]]
 
     def __init__(
         self,
-        value: BaseValue[TPayload],
+        value: Union[
+            BaseValue[Union[FilteredPayload, TPayload]],
+            BaseValue[FilteredPayload],
+            BaseValue[TPayload],
```
fpacifici, line 56: Not sure I get why we need all three cases.

fpacifici (commented):
I think we may be looking at this problem from the wrong angle.
Most customer defined strategies are not supposed to care about filtered messages and should just let them pass through. 
If we are requiring every strategy to take them into account by default, I feel we are optimizing for the wrong goal.
In order to make it easy to do the right thing we should probably try to make ignoring filtered messages a noop that the developer does not have to think about. That reduces cognitive overload when using the library. 

Today, when creating a new strategy, we inherit ProcessingStrategy, which is just an interface. What if we added the logic that sends the fitlered messages through in the strategy we inherit from? (Or in alternative, what if we wrap the custom strategy with a wrapper that takes care of passing filtered messages through) ? 

Basically the type of the message used in a custom strategy would not include the fitlered message by default. The wrapping strategy would take care of that type and pass to the custom logic only valid messages that are not filtered. 

Example:
```
class FilteringStrategy(ABC, ProcessingStrategy[BaseValue[FilterableTPayload]) # types would have to change here.
# This can receive FitleredMessage.
def __init__(self, inenr_strategy: ProcessingStrategy[TPayload]):
    # inner strategy does not take fitlered paylaods into account.
    ...
def submit (...):
    if isinstance(message, FitleredMessage):
        #pass it through to the next step.
    self.inner_strategy(message.actual_value)
```

untitaker:
>  Most customer defined strategies are not supposed to care about filtered messages and should just let them pass through.

I think the way we're approaching this is that we assume most people who would want to write a custom strategy can achieve their goals by using a combination of RunTask and other strategies.

In other words, custom strategy classes are no longer really the #1 way to run custom processing logic.

> `class FilteringStrategy(ABC, ProcessingStrategy[BaseValue[FilterableTPayload]) # types would have to change here.`

Jumping on the "the types would have to change here" comment a bit....

The initial version of this was indeed one where we have two separate concrete types for strategies that can handle filtered messages vs ones that can't:

```
class StrategyThatCanFilter(ABC, ProcessingStrategy[Union[FilteredPayload, T]])
class StrategyThatCantFilter(ABC, ProcessingStrategy[T])
```

then you could introduce your proposed wrapper strategy to turn instances of the latter into instances of the former. (i didn't think that far though)

arroyo/types.py
untitaker: I added a code comment

arroyo/processing/strategies/batching.py
untitaker: I assume `BatchStep` and `UnbatchStep` always come in pairs. `BatchStep` should never produce a `FilteredMessage`.

I just realized however one could theoretically put a filter step between both of them, so the outer message could be filtered.

I updated the strategy to deal with filtered messages "normally"

arroyo/processing/strategies/filter.py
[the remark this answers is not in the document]
lynnagara: How about we remove any offsets staged in `self.__uncommitted_offsets` for that partition if the `test_function` passes and we have a message to actually send through the pipeline. That way, we can safely throttle flushing the uncommitted offsets so it actually respects the commit policy and we won't get offsets committed out of order. The rate of filtered messages being forwarded would be the same regardless of how many messages in between actually do pass the filtering.

lynnagara:
How many of the strategies we have in Sentry/Snuba today would need to be updated to manually handle the new type  Would it only be those that actually use the FilterStep or others as well?

fpacifici:
> Would it only be those that actually use the FilterStep or others as well?

@lynnagara, if the strategy `submit` method has to take a `Union[FilteredSomething, Something]` then the strategy implementation will have to take both into account whether or not the consumer has a filter at the beginning.
Assuming developers would know when to ignore the existence of the filter step seems an api design that encourages an unsafe behavior: ignore one of the types you can receive.

> The initial version of this was indeed one where we have two separate concrete types for strategies that can handle filtered messages vs ones that can't:

I think at a high level I like that approach (at the condition that the `StrategyThatCantFilter` is wrapped in one that can take care of the filter). What was the reason for abandoning it ?

Basically I would focus on keeping simple for the developer and reduce their cognitive load. If the code of their strategy does not to know about filtered messages it should not know or care about them (not by ignoring one of the types the submit function can receive). Removing the cognitive overhead from Kafka usage was the initial goal of Arroyo.

untitaker:
>I think at a high level I like that approach (at the condition that the StrategyThatCantFilter is wrapped in one that can take care of the filter). What was the reason for abandoning it ?

It did mean that every existing strategy had to be changed to this:

```
class MyStrat(ProcessingStrategy[Union[FilteredPayload, RealPayload]]]): ...
```

all messages passed to that strategy are of type `Message[Union[FilteredPayload, RealPayload]]`.

A user-defined strategy that cannot deal with filtered payloads would take a type `Message[RealPayload]`, and therefore subclass from `ProcessingStrategy[RealPayload]` (as it already does).

The real issues start with the contravariance of `TPayload` (or lack thereof). MyStrat[Union[FilteredPayload, RealPayload]]` should be cast-able to `MyStrat[RealPayload]` (contravariance), however, `Message[RealPayload]` should be cast-able in the other direction, to `Message[Union[FilteredPayload, RealPayload]]`

So `TPayload` would now have to be split up into two types: One that is covariant, and another that is contravariant.

It was suggested to me that instead, we hide the `Union[FilteredPayload, ..]` in `Message`, and a natural consequence of that to me is that every strategy taking any sort of `Message` needs to deal with filtered payloads.

Since `TPayload` is still one (invariant) type, we need some casts here and there.

arroyo/processing/strategies/reduce.py
```diff
@@ -115,6 +115,10 @@ def submit(self, message: Message[TPayload]) -> None:
         """
         assert not self.__closed
 
+        if isinstance(message.payload, FilteredPayload):
```
untitaker, line 118: I discovered this is broken -- it forwards messages to the next step out of order, which causes further problems down the road

untitaker:
I have submitted an alternative PR at https://github.com/getsentry/arroyo/pull/185

untitaker:
closed in favor of https://github.com/getsentry/arroyo/pull/185

PR #179, opened 2023-01-30
ref: upgrade isort to work around poetry breakage 

asottile-sentry:
https://github.com/python-poetry/poetry-core/pull/542

Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
         args: [--config-file, mypy.ini, --strict]
         additional_dependencies: [pytest==7.1.2]
   - repo: https://github.com/pycqa/isort
-    rev: 5.8.0
+    rev: 5.12.0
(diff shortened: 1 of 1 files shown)


evanh (approved):

asottile-sentry:
@evanh I can't seem to merge here -- is the repo set to require up-to-date branches?

evanh:
Yes it does. Should we remove that?

asottile-sentry:
that's what I'd recommend heh -- you can check the box that *allows* that dropdown to be there, but I find requiring it is a really bad experience (I shouldn't have to rebase every open PR because someone fixed a typo for example)

evanh:
OK well @lynnagara is the owner of this repo. I'll let her make the call.

lynnagara:
Disabled the rule, it's mergeable without updating now.

PR #180, opened 2023-02-14
ref: Add more timings to decoder.json, add non-strict validation mode

untitaker:
I want this to figure out how fast validation vs JSON decoding alone is. The non-strict mode is for avoiding consumer crashes while the schema is not entirely fleshed out yet.

### DIFF
--- a/arroyo/processing/strategies/decoder/json.py
+++ b/arroyo/processing/strategies/decoder/json.py
@@ -1,5 +1,5 @@
 import time
-from typing import Optional
+from typing import Mapping, Optional
 
 import fastjsonschema  # pip install sentry-arroyo[json]
 import rapidjson
@@ -9,8 +9,17 @@ from arroyo.utils.metrics import get_metrics
 
 
 class JsonCodec(Codec[object]):
-    def __init__(self, json_schema: Optional[object] = None) -> None:
+    def __init__(
+        self,
+        json_schema: Optional[object] = None,
+        json_schema_name: Optional[str] = None,
+    ) -> None:
         self.__metrics = get_metrics()
+        self.__metrics_tags: Mapping[str, str]
+        if json_schema_name:
+            self.__metrics_tags = {"schema_name": json_schema_name}
+        else:
+            self.__metrics_tags = {}
 
         if json_schema is not None:
             self.__validate = fastjsonschema.compile(json_schema)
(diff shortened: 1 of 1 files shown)


evanh (approved):

arroyo/processing/strategies/decoder/json.py
```diff
@@ -1,26 +1,72 @@
...
+
+logger = logging.getLogger(__name__)
 
 
 class JsonCodec(Codec[object]):
-    def __init__(self, json_schema: Optional[object] = None) -> None:
+    def __init__(
+        self,
+        json_schema: Optional[object] = None,
+        json_schema_name: Optional[str] = None,
```
lynnagara, line 18: nit: I'd skip this. We already tag the metrics backend with the storage/consumer group/topic before passing it to Arroyo, so I don't think there's too much ambiguity about which consumer any metrics are coming from.

arroyo/processing/strategies/decoder/json.py
```diff
@@ -1,26 +1,72 @@
...
+            "arroyo.strategies.decoder.json.loads",
+            after_decoded,
+            tags=self.__metrics_tags,
+        )
+
         if validate:
             self.validate(decoded)
+            self.__metrics.timing(
+                "arroyo.strategies.decoder.json.validate",
+                time.time() - after_decoded,
```
lynnagara, line 52: I think this metric is off since `after_decoded` subtracts the start time

lynnagara (commented):
Can we throttle/buffer/sample rather than emitting metrics on every message? Similar to what we do in the processor in https://github.com/getsentry/arroyo/blob/main/arroyo/processing/processor.py#L22-L55

arroyo/processing/strategies/decoder/json.py
```diff
@@ -1,26 +1,63 @@
...
     def decode(self, raw_data: bytes, validate: bool) -> object:
+        start = time.time()
         decoded = rapidjson.loads(raw_data)
+
+        after_decoded = time.time()
+
+        self.__metrics.timing(
+            "arroyo.strategies.decoder.json.loads",
+            after_decoded - start,
+        )
```
lynnagara, line 39: This is emitting a metric on every message, which can affect performance. Can we find a way to send this metric say once per second or so?

untitaker: can't think of a better place to emit this metric, or to do it in batches, so I just put this behind sampling.

untitaker: oh, sorry, i just saw the other message. will use that

untitaker:
Upon taking a closer look at the linked metrics buffer, I decided that this is just no good place to emit metrics at all. The loop is too hot. We can emit metrics much more easily and efficiently in the consumer where we do batching.

PR #189, opened 2023-02-23
fix: Fix multiprocessing join

lynnagara:
We weren't catching the TimeoutError, leading to consumer to crash during join anytime the timeout is hit.


untitaker (commented):
doesn't that make us silently lose messages?

lynnagara:
The purpose of the join timeout is to stop processing messages when the deadline is hit and shut down the strategy. Since these messages will never get committed they'll get reprocessed when the strategy is re-created.

untitaker (approved):
nevermind, i thought `join()` was supposed to raise on timeout

PR #193, opened 2023-03-10
feat: Add ability to assign consumer to a subset of partitions

lynnagara:
Today Arroyo's consumer only works with auto assignment, and doesn't support a consumer getting assigned a subset of partitions and offsets.

This change brings it closer to how the assignment callback works in the actual Kafka consumer API where auto assignment only happens if an assignment callback is not passed, otherwise the user should manually assign in the callback.

The immediate motivation for this change is to support a dead letter queue consumer implementation which will seek to specific partitions/offsets and fetch the invalid messages from those positions.

### DIFF
--- a/arroyo/backends/abstract.py
+++ b/arroyo/backends/abstract.py
@@ -121,6 +121,14 @@ class Consumer(Generic[TStrategyPayload], ABC):
         raise NotImplementedError
 
     @abstractmethod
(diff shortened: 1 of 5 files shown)


lynnagara:
Not needed for DLQ, we are going with an approach that buffers messages in stream processor instead

PR #197, opened 2023-03-11
feat: Add unfold strategy

lynnagara:
This strategy is the opposite of Reduce. The objective is to provide a more flexible way to extract a number of messages from a single message than UnbatchStep currently allows. With UnbatchStep, the input message must be `ValuesBatch`, which is an array of BaseValue from the BatchStep. Unfold allows for any user defined type and user provided generation function.

This change is motivated by a recent attempt to use Batch/Unbatch in the Sentry indexer (https://github.com/getsentry/sentry/pull/44721). The other strategies in the pipeline expected a slightly different format to what the BatchStep outputs. Switching out single steps in this pipeline without changing the rest would be easier if we were able to define both custom fold and unfold functions.

### DIFF
--- a/arroyo/processing/strategies/unfold.py
+++ b/arroyo/processing/strategies/unfold.py
@@ -1,0 +1,104 @@
+import time
+from collections import deque
+from typing import Callable, Collection, Deque, Generic, Optional, TypeVar, Union, cast
+
+from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
+from arroyo.types import FilteredPayload, Message, Value
+
+TInput = TypeVar("TInput")
+TOutput = TypeVar("TOutput")
+
+
+class Unfold(
+    ProcessingStrategy[Union[FilteredPayload, TInput]], Generic[TInput, TOutput]
+):
+    """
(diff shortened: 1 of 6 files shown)


arroyo/processing/strategies/unfold.py
```diff
@@ -0,0 +1,104 @@
...
+        # If we get MessageRejected from the next step, we put the pending messages here
+        self.__pending: Deque[Message[TOutput]] = deque()
+
+    def submit(self, message: Message[Union[FilteredPayload, TInput]]) -> None:
+        assert not self.__closed
+        if self.__pending:
+            raise MessageRejected
+
+        if isinstance(message.payload, FilteredPayload):
+            self.__next_step.submit(
```
untitaker, line 45: this reorders messages, so you can't do this. imagine three messages where the second one is filtered. it will be sorted in front of the other ones

untitaker (changes requested):

untitaker (approved):

arroyo/processing/strategies/unfold.py
untitaker: oh right...

PR #201, opened 2023-03-23
ref: Make stream processor / producer / consumer not generic

lynnagara:
Throwing this up as an alternative idea to https://github.com/getsentry/arroyo/pull/200.

I don't think there's any reason why stream processor, producer and consumer actually need to be generic. It makes typing more annoying for users and AFAIK there's no reason why we shouldn't enforce that a message consumed or produced needs to be a KafkaPayload - that is - a tuple of key, value (bytes) and headers.

### DIFF
--- a/arroyo/backends/abstract.py
+++ b/arroyo/backends/abstract.py
@@ -3,14 +3,14 @@ from __future__ import annotations
 import logging
 from abc import ABC, abstractmethod, abstractproperty
 from concurrent.futures import Future
-from typing import Callable, Generic, Mapping, Optional, Sequence, Union
+from typing import Callable, Mapping, Optional, Sequence, Union
 
-from arroyo.types import BrokerValue, Partition, Topic, TStrategyPayload
+from arroyo.types import BrokerValue, KafkaPayload, Partition, Topic
 
 logger = logging.getLogger(__name__)
 
 
-class Consumer(Generic[TStrategyPayload], ABC):
+class Consumer(ABC):
     """
     This abstract class provides an interface for consuming messages from a
     multiplexed collection of partitioned topic streams.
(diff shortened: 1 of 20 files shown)


arroyo/backends/abstract.py
```diff
@@ -64,7 +64,7 @@ def unsubscribe(self) -> None:
     @abstractmethod
     def poll(
         self, timeout: Optional[float] = None
-    ) -> Optional[BrokerValue[TStrategyPayload]]:
+    ) -> Optional[BrokerValue[KafkaPayload]]:
```
lynnagara, line 67: Also considered renaming this into just `Payload` and keeping `KafkaPayload` just as an alias

untitaker:
I believe that this is fine for as long as it is easy to construct KafkaPayload in tests. Being generic over the payload type has the advantage that one can do either one of those two things:

1. hold onto rdkafka specific types from within KafkaPayload, which could make sense to avoid copying bytes around from C++-land to Python land
2. test streaming consumer without testing KafkaPayload

currently i suppose we do neither

however, in Rust-land we do use rdkafka-types in KafkaPayload: https://github.com/getsentry/snuba/blob/be3090edff8f83697bd09fe4d68a4ed4efc16614/rust_snuba/rust_arroyo/src/backends/kafka/types.rs#L3-L8

So I think if we translated this design to Rust, it would become harder to test StreamProcessor

lynnagara:
I guess I don't quite get why having Payload needs to be generic to achieve this goal... Can't we evolve KafkaPayload in the future to hold the pointers? The `KafkaConsumer` and `KafkaProducer` already not generic anyway - only the local version used for testing that actually is.

untitaker:
I'm a bit fuzzy how all those different consumer/processor classes compose, I mostly just care that functions that took TPayload and now take KafkaPayload are still sort of unittestable without actually talking to kafka, if that is provided I guess we're good.

lynnagara:
Let's figure out how this might work in Rust first before making a decision here

PR #207, opened 2023-04-04
feat: Make DLQ buffer size configurable

lynnagara:
If commit isn't being called regularly, it may grow too large, let users configure the buffer size.

### DIFF
--- a/arroyo/dlq.py
+++ b/arroyo/dlq.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


untitaker (approved):
what is our plan if we want to dlq a message outside of the buffer? crashloop the consumer and hope we have more luck next time?

lynnagara:
> what is our plan if we want to dlq a message outside of the buffer? crashloop the consumer and hope we have more luck next time?

Yeah, that's pretty much it. But I figure this is a more rare case where the consumer is misconfigured AND there's an invalid message.

PR #209, opened 2023-04-05
fix: Prevent repeated calls to consumer.pause() if already paused

lynnagara:
Let's prevent excessive calls to pause if we are already paused anyway.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -258,7 +258,9 @@ class StreamProcessor(Generic[TStrategyPayload]):
                             e,
                             self.__message,
                         )
-                        self.__consumer.pause([*self.__consumer.tell().keys()])
+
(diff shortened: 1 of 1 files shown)


arroyo/processing/processor.py
```diff
@@ -258,7 +258,9 @@ def _run_once(self) -> None:
                             e,
                             self.__message,
                         )
-                        self.__consumer.pause([*self.__consumer.tell().keys()])
+
+                        if self.__paused_timestamp is not None:
```
onewland, line 262: should that be `is None`?

lynnagara: Oops, you are right. Thanks

onewland (approved):

fpacifici (approved):

PR #211, opened 2023-04-12
fix: Fix installation of sphinx-autodoc-typehints

untitaker:
Seems like the extra got randomly renamed in an effort to modernize
packaging: https://github.com/tox-dev/sphinx-autodoc-typehints/issues/263

Also relax requirements on sphinx so that there's no dep conflict


lynnagara:
@hubertdeng123 can you help with this failure?

lynnagara (approved):

PR #212, opened 2023-04-14
fix: Do not enter invalid state when strategy.submit raises InvalidMessage

untitaker:


arroyo/processing/processor.py
```diff
@@ -230,9 +239,17 @@ def _run_once(self) -> None:
...
-            # If a message was carried over from the previous run, the consumer
-            # should be paused and not returning any messages on ``poll``.
-            if self.__consumer.poll(timeout=0) is not None:
+            # If a message was carried over from the previous run, there are two reasons:
+            #
+            # * MessageRejected. the consumer should be paused and not
+            #   returning any messages on ``poll``.
+            # * InvalidMessage. the message should be resubmitted.
+            #   _handle_invalid_message is responsible for clearing out
+            #   self.__message if it was the invalid one.
```
lynnagara, line 248: Thanks, this helps a lot

lynnagara: There is a comment above that states `Do not "carry over" message if it is the invalid one` - which kind of implies an invalid message is never actually carried over

untitaker: if InvalidMessage is raised, it may or may not relate to the message that is currently assigned to `self.__message`. If it is the same message, it is not carried over. If it isn't, the message is carried over.

untitaker: so specifically what can also happen in theory, is that `submit(message)` raises InvalidMessage, but that exception was for a previous message. In that case we want to reattempt the submission of `message`.

PR #214, opened 2023-04-17
feat: Measure time to join strategy [SNS-2154]

untitaker:


arroyo/processing/processor.py
```diff
@@ -111,10 +112,18 @@ def _close_strategy() -> None:
...
                 try:
                     self.__processing_strategy.join(self.__join_timeout)
+                    self.__metrics_buffer.increment(
+                        ConsumerTiming.CONSUMER_JOIN_TIME, time.time() - start_join
+                    )
                     break
                 except InvalidMessage as e:
+                    self.__metrics_buffer.increment(
+                        ConsumerTiming.CONSUMER_JOIN_TIME, time.time() - start_join
+                    )
```
lynnagara, line 126: should this be below line 127?

untitaker: if we do that, then CONSUMER_DLQ_TIME overlaps with CONSUMER_JOIN_TIME

lynnagara: Yeah. I'm worried the join time is understated if we don't. Knowing the actual total join time is important here, not just the ratio of join time to other time (as we want to know how close to the 5 second mark - when we start getting those rebalance timeout errors - we are).

lynnagara: could be CONSUMER_JOIN_DLQ_TIME?

untitaker: How about CONSUMER_SHUTDOWN_TIME is added as additional measurement? #215 already adds one that overlaps with join/poll timings

lynnagara: Ignore that, separate CONSUMER_SHUTDOWN_TIME is better

lynnagara (approved):

PR #215, opened 2023-04-17
feat: Add error instrumentation and timing to arroyo callbacks

untitaker:


arroyo/processing/processor.py
```diff
@@ -22,6 +33,28 @@
...
+
+def _rdkafka_callback(metrics: MetricsBuffer) -> Callable[[F], F]:
+    def decorator(f: F) -> F:
+        @functools.wraps(f)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            start_time = time.time()
+            try:
+                return f(*args, **kwargs)
+            except Exception:
+                logger.exception(f"{f.__name__} crashed")
```
nikhars, line 47: Is it necessary to log an exception in the decorator? Would it be better if you just raise the exception and let the caller take care of whether to log it or not?

For example, if you log it here, would [this](https://github.com/getsentry/arroyo/pull/215/files#diff-57203c74e6cc151b8eef17b234894b09b11d21af20f82ab2d9cf8181150a9deaR179) code also log the exception? That would lead to 2 logs with exceptions.

untitaker: That other code will not log the exception because at that time there is no exception on `sys.exc_info`. I believe that `logger.exception` there was a mistake though, I don't see any chance that this does anything different from `logger.error` in that particular situation. We should change it there imo

> Would it be better if you just raise the exception and let the caller take care of whether to log it or not?

Right now it seems that the caller, rdkafka, does not log those exceptions properly. During rollout of DLQ we found that a lot of crashes during partition reassignment are not visible in Sentry.

The Sentry SDK will deduplicate exception objects by identity. `if exc1 is exc2`, only one will be reported to Sentry.

PR #217, opened 2023-04-26
fix(dlq): Fix crash during join when InvalidMessage in flight

lynnagara:
During the DLQ test, we hit the `assert not self.__closed` error during rebalancing as we were attempting to submit filtered payload messages to strategies that we already closed.

This change moves the `next_step.submit(filtered_payload)` call before the `next_step.close()` call to avoid this bug.

If any strategy raises InvalidMessage during join, the strategy guard will drop it - this is safe to do as it will be reprocessed after the partition is reassigned and the invalid message should only get produced and committed once.


untitaker (approved):
hmm, can we add a test for that?

lynnagara:
Oops, merged before seeing the comment about test. Will open another PR for it.

PR #218, opened 2023-04-26
fix: Fix RunTaskInThreads to handle InvalidMessage during close, add tests

lynnagara:
Follow up to https://github.com/getsentry/arroyo/pull/217. Ports the same fix for `RunTaskInThreads`.

Adds a test for `RunTask`, `RunTaskInThreads`, and `RunTaskWithMultiprocessing`.

### DIFF
--- a/arroyo/processing/strategies/run_task_in_threads.py
+++ b/arroyo/processing/strategies/run_task_in_threads.py
@@ -140,12 +140,12 @@ class RunTaskInThreads(
             self.__next_step.poll()
             self.__next_step.submit(next_message)
 
+        self.__next_step.close()
         self.__executor.shutdown()
         self.__next_step.join(timeout)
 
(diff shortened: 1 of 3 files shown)


evanh (approved):

untitaker (changes requested):
i believe this test should be in `test_all`. You're just testing the guard strategy here, but the guard strategy is not something that _wraps_ all strategies

untitaker:
it may also be that some strategies that do not use guard still have the bug that #217 was supposed to fix?

lynnagara:
> it may also be that some strategies that do not use guard still have the bug that #217 was supposed to fix?

You are right.... RunTaskInThreads was also not handled correctly. Updated with the fix for that strategy and the test for all of the run task family of strategies.

PR #219, opened 2023-05-02
ref: Remove extraneous codecs from arroyo

untitaker:
As of version 0.1.0, sentry-kafka-schemas now contains this code.

Arroyo still needs to have some concept of encoding because it
implements commit log topics... Not entirely sure what to do there. I
guess we'll just keep a minimal codecs impl around?


lynnagara:
I'd like to eventually move the commit log codec to `sentry-kafka-schemas` too. It's bugged me for a while that it was in Arroyo as it's extremely specific to how Sentry uses that topic. Wdyt about changing the format to json while we're at it rather than relying on information spread between the header, key and value that has to be decoded in a very specific way?

setup.py
```diff
@@ -30,8 +30,6 @@ def get_requirements() -> Sequence[str]:
         "Topic :: Software Development :: Libraries :: Python Modules",
     ],
     extras_require={
-        "avro": ["avro>=1.11.1"],
         "json": ["python-rapidjson>=1.8", "fastjsonschema>=2.16.2"],
```
lynnagara, line 33: Do we wanna remove this too?

untitaker: but commit log codec needs it, right?

untitaker: ah oops indeed it just uses the Codec class. should we remove that one aswell?

lynnagara: Yeah I think we should since it's in sentry-kafka-schemas (IIRC it's a different one to the commit log codec)

lynnagara (approved):

PR #220, opened 2023-05-03
fix: Add metrics around RunTaskWithMultiprocessing + docs [SNS-2204]

untitaker:
I am investigating an issue where increasing the buffer size increased consumer lag, e2e latency metrics and avg batch size. Turns out that this behavior is intentional, and we just have to document it better and add instrumentation.

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -458,7 +458,7 @@ class RunTaskWithMultiprocessing(
                 result.next_index_to_process = idx
 
         if result.next_index_to_process != len(input_batch):
-            self.__metrics.increment("batch.output.overflow")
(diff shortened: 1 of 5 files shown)


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -421,6 +458,7 @@ def __check_for_results(self, timeout: Optional[float] = None) -> None:
                 result.next_index_to_process = idx
 
         if result.next_index_to_process != len(input_batch):
+            self.__metrics.increment("batch.output.overflow")
```
lynnagara, line 461:
```suggestion
            self.__metrics.increment("arroyo.strategies.run_task_with_multiprocessing.batch.output.overflow")
```
It's a mouthful I know but I've started prefixing new metrics this way, and would like to migrate the existing ones at some point

lynnagara (approved):

PR #221, opened 2023-05-03
ref: Clean up documentation

untitaker:
while working on https://github.com/getsentry/arroyo/pull/220:

* swap out the docs theme for a nicer one
* remove a bunch of autogenerated documentation that isn't useful to
  arroyo users (join(), close(), MessageBatch)
* remove documentation for legacy DLQs, decoders, encoders and
  TransformStep. Codecs kind of moved into sentry-kafka-schemas,
  decoders and encoders ended up not being used (and also reference
  arroyo's own codecs), transform step is an alias for RunTask.
* Fix broken mermaid diagrams by upgrading the library


lynnagara (approved):
awesome, thanks!

PR #236, opened 2023-05-18
docs: Minor readability details

kamilogorek:
Spacing, invalid code blocks, unified indents, added `timeout=` to make it obvious what that `0` actually is.

### DIFF
--- a/docs/source/what_for.rst
+++ b/docs/source/what_for.rst
@@ -53,9 +53,7 @@ The most simple kafka consumer looks something like this:
 
    while True:
        message = consumer.poll()
-       send_to_destination(
-           process_message(message)
-     )
+       send_to_destination(process_message(message))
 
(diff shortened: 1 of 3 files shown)


docs/source/what_for.rst
```diff
@@ -232,7 +230,7 @@ librdkafka’s Callback Hell
 
 librdkafka uses callbacks as a core mechanic for control flow. A few
 such examples have been mentioned in this document already. What is not
-clear however, is that **callbacks are only called when ``poll`` is
+clear however, is that **callbacks are only called when `poll` is
```
untitaker, line 233:
```suggestion
clear however, is that **callbacks are only called when ``poll`` is
```

it's not markdown (unfortunately), so `<code>` means two ticks

untitaker:
ty!

PR #237, opened 2023-05-18
fix(multiprocessing): Honor join() timeout even if processing pool is overloaded [INC-378]

untitaker:
`Pool.join` will block for more than `timeout` seconds if there is a task that takes more than `timeout` seconds. We can just terminate the pool though, because we have effectively reimplemented `join()` ourselves using `__check_for_results`.

Note that when I say "task", I mean "task" from the POV of `multiprocessing`, which translates to a call to `apply_async`, which is an entire message batch from arroyo POV

This effect is potentially made worse if `next_step` is rejecting all messages:

* When running `git checkout origin/main arroyo/`, the new tests hang forever in join, because the batch keeps getting reprocessed
* When additinally removing `side_effect = MessageRejected()` from the new tests, they fail because `join()` "only" terminates whenever the pool is drained, which is 10 seconds.

The fact that we reprocess the remaining batch on backpressure is kinda dumb anyway, let's fix it in #238

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -614,8 +614,6 @@ class RunTaskWithMultiprocessing(
             timeout=timeout,
         )
 
-        self.__pool.close()
-
         logger.debug("Waiting for %s...", self.__pool)
         self.__pool.terminate()
 
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -431,6 +431,36 @@ class RunTaskWithMultiprocessing(
                     pass
 
     def __check_for_results(self, timeout: Optional[float] = None) -> None:
+        deadline = time.time() + timeout if timeout is not None else None
+
+        while self.__processes:
+            try:
+                self.__check_for_results_impl(
+                    timeout=max(deadline - time.time(), 0)
+                    if deadline is not None
+                    else None
+                )
+            except multiprocessing.TimeoutError:
+                if self.__pool_waiting_time is None:
+                    self.__pool_waiting_time = time.time()
+                else:
+                    current_time = time.time()
+                    if current_time - self.__pool_waiting_time > LOG_THRESHOLD_TIME:
+                        logger.warning(
+                            "Waited on the process pool longer than %d seconds. Waiting for %d results. Pool: %r",
+                            LOG_THRESHOLD_TIME,
+                            len(self.__processes),
+                            self.__pool,
+                        )
+                        self.__pool_waiting_time = current_time
+                break
+            else:
+                self.__pool_waiting_time = None
+
+                if deadline is not None and time.time() > deadline:
+                    break
+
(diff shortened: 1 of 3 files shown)


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -601,23 +610,14 @@ def join(self, timeout: Optional[float] = None) -> None:
...
-            except multiprocessing.TimeoutError:
-                pass
+        self.__check_for_results(
+            timeout=timeout,
+        )
 
         self.__pool.close()
 
         logger.debug("Waiting for %s...", self.__pool)
-        # ``Pool.join`` doesn't accept a timeout (?!) but this really shouldn't
```
onewland, line 617: "shouldn't"

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -601,23 +610,14 @@ def join(self, timeout: Optional[float] = None) -> None:
...
-        # block for any significant amount of time unless something really went
-        # wrong (i.e. we lost track of a task)
-        self.__pool.join()
+        self.__pool.terminate()
```
lynnagara, line 620: Is this change actually needed now that RunTaskWithMultiprocessing.join() is applying the proper timeout in the check for results step? I think it's better to shut down cleanly with `pool.close()` + `pool.join()` rather than `terminate()`.

untitaker: if you do that and submit a task that takes 10 seconds, and call `join(timeout=3)`, `__check_for_results` will return after 3 seconds. then, `pool.join()` will block for another 7 seconds because the task is still not done

lynnagara: Ok. Is there a reason for keeping the `pool.close()` call directly before this?

untitaker: Ah, no. I also wonder if we should block on the pool shutting down by calling terminate + join instead of just terminate. I am not sure we should really block rebalance on it

lynnagara: `terminate()` makes sense to me if we have exceeded the deadline already. Though if no timeout is passed, I think it's more correct to honor that and close + join instead of terminate.

untitaker: by the time we call `terminate()`, the deadline is either guaranteed to have passed, or the pool is guaranteed to be empty.

If no deadline is set, `__check_for_results` will block until the pool is empty. So if we `join` after that, it can only either block for (much) longer than the deadline allows, or do nothing useful at all

lynnagara: Ah I see, terminate makes sense then. Still, why do we call `pool.close()` before `pool.terminate()`?

lynnagara (approved):

arroyo/processing/strategies/run_task_with_multiprocessing.py
untitaker: Oh yeah I forgot about that, will fix on Monday

PR #238, opened 2023-05-18
fix(multiprocessing): More performant backpressure [INC-378]

untitaker:
Do not resubmit a batch when next_step raises MessageRejected. We can easily implement that now because we had to do something similar for DLQing

Batches are still re-serialized when output buffer overflows, so that can still have catastrophic perf impact.

Depends on a refactoring made in #237 

## Why as part of incident response?

There is no strong evidence that this is the culprit of the problems we have in INC-378, such as lack of CPU saturation. However, we know two things:

* We are constantly dissatisfied with throughput of our consumers
* Due to recently added metrics, we know that snuba generic-metrics keeps encountering backpressure from http step towards multiprocessing step

Because of (2), IMO this performance bottleneck must be a major contributor to (1)

If there is continuous MessageRejected, it could be that we process as much as every second batch a second time (and then maybe a third time). `n + (n - 1) + ... + 1` is still `O(n^2)`

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -445,6 +445,8 @@ class RunTaskWithMultiprocessing(
                     else None
                 )
             except NextStepTimeoutError:
+                if deadline is None or deadline > time.time():
+                    continue
                 break
             except multiprocessing.TimeoutError:
                 if self.__pool_waiting_time is None:
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -431,6 +431,36 @@ class RunTaskWithMultiprocessing(
                     pass
 
     def __check_for_results(self, timeout: Optional[float] = None) -> None:
+        deadline = time.time() + timeout if timeout is not None else None
+
+        while self.__processes:
+            try:
+                self.__check_for_results_impl(
+                    timeout=max(deadline - time.time(), 0)
+                    if deadline is not None
+                    else None
+                )
+            except multiprocessing.TimeoutError:
+                if self.__pool_waiting_time is None:
+                    self.__pool_waiting_time = time.time()
+                else:
+                    current_time = time.time()
+                    if current_time - self.__pool_waiting_time > LOG_THRESHOLD_TIME:
+                        logger.warning(
+                            "Waited on the process pool longer than %d seconds. Waiting for %d results. Pool: %r",
+                            LOG_THRESHOLD_TIME,
+                            len(self.__processes),
+                            self.__pool,
+                        )
+                        self.__pool_waiting_time = current_time
+                break
+            else:
+                self.__pool_waiting_time = None
+
+                if deadline is not None and time.time() > deadline:
+                    break
+
+    def __check_for_results_impl(self, timeout: Optional[float] = None) -> None:
         input_batch, async_result = self.__processes[0]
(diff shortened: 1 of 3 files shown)


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -483,24 +488,29 @@ def __check_for_results_impl(self, timeout: Optional[float] = None) -> None:
...
-            else:
-                self.__metrics.increment(
-                    "arroyo.strategies.run_task_with_multiprocessing.batch.output.overflow"
+
+                # This is a warning because backpressure on the multiprocessing
+                # strategy _might_ cause degraded CPU saturation.
+                logger.warning(
+                    "Received backpressure from next_step, splitting output batch (%0.2f%% complete)",
+                    idx / len(input_batch) * 100,
                 )
```
lynnagara, line 503: What do you think about omitting this warning entirely? IMO it's a bit too noisy and the backpressure metric on the stream processor seems like it's enough. None of the other strategies emit something similar to this.

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -440,6 +444,8 @@ def __check_for_results(self, timeout: Optional[float] = None) -> None:
                     if deadline is not None
                     else None
                 )
+            except NextStepTimeoutError:
+                break
```
lynnagara, line 448: Should this be `continue`? So that we continue checking for results until the timeout is reached. Otherwise we may be giving up before the timeout is reached.

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -483,24 +488,29 @@ def __check_for_results_impl(self, timeout: Optional[float] = None) -> None:
...
-                self.__metrics.increment(
-                    "arroyo.strategies.run_task_with_multiprocessing.batch.output.overflow"
+
+                # This is a warning because backpressure on the multiprocessing
+                # strategy _might_ cause degraded CPU saturation.
+                logger.warning(
+                    "Received backpressure from next_step, splitting output batch (%0.2f%% complete)",
+                    idx / len(input_batch) * 100,
                 )
+                raise NextStepTimeoutError()
```
nikhars, line 504: Would it be beneficial to add details around how much of the batch was processed before having to split it in the exception itself?

arroyo/processing/strategies/run_task_with_multiprocessing.py
untitaker: ok makes sense, removed

arroyo/processing/strategies/run_task_with_multiprocessing.py
untitaker: it would mean potentially spinning here. I would rather re-raise MessageRejected so the spinning happens in processor only, but I haven't thought it through, so I'd rather not...

arroyo/processing/strategies/run_task_with_multiprocessing.py
untitaker: it is logged in the warning, but lyn argues that we should have no warning. at the same time this exception is caught in the outer loop, it is effectively only for control flow

arroyo/processing/strategies/run_task_with_multiprocessing.py
untitaker: You are right, this is a bug, fixed.

lynnagara (approved):

PR #239, opened 2023-05-18
feat: Expose run_once method on the stream processor

lynnagara:
When processing DLQ messages we want the ability to stop after a certain number of messages. In order to support this feature, we expose the `StreamProcessor.run_once()` method, which can be called in a loop outside of the stream processor class.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -236,6 +236,14 @@ class StreamProcessor(Generic[TStrategyPayload]):
             )
             self.__commit_policy_state.did_commit(now, offsets)
 
+    def run_once(self) -> None:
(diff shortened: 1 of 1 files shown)


untitaker:
can't we track offsets in a custom strategy? it seems to be more correct to track offsets rather than how often stream processor looped

lynnagara:
Yeah, this doesn't actually track offsets at all since there's no guarantee a message was available to process. Will look into the other option. But we will need a way to terminate the consumer from inside a strategy then?

untitaker:
i think we can call `sys.exit`, or otherwise we can raise a custom exception and catch it from the outside

PR #241, opened 2023-05-23
ref(multiprocessing): Rename some metrics and add documentation

untitaker:


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -299,21 +299,61 @@ class RunTaskWithMultiprocessing(
...
+    amortized with larger batches. ``max_batch_size`` and ``max_batch_time``
+    should be tweaked for optimal performance. You can observe the effect in the following metrics:
+
+    * ``arroyo.strategies.run_task_with_multiprocessing.batch.size.msg``: The number of messages per batch.
+    * ``arroyo.strategies.run_task_with_multiprocessing.batch.size.bytes``: The number of bytes used per batch.
+
+    If ``batch.size.msg`` is flat (as in, it's a perfectly straight line at a
+    constant), you are hitting ``max_batch_size``. If ``batch.size.bytes`` is
+    flat, you are hitting input buffer overflow (see next section). If neither
+    are flat, you are hitting batch timeout.
```
untitaker, line 331:
```suggestion
    are flat, you are hitting ``max_batch_time``.
```

untitaker:
in Relay we have a central registry of metric names (basically a big global enum instead of strings). If we had that in arroyo, we could automatically generate docs from it

lynnagara:
> in Relay we have a central registry of metric names (basically a big global enum instead of strings). If we had that in arroyo, we could automatically generate docs from it

This seems doable

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -299,21 +301,77 @@ class RunTaskWithMultiprocessing(
...
+
+    You want to keep an eye on these metrics:
+
+    1. ``arroyo.strategies.run_task_with_multiprocessing.batch.input.overflow``
+    2. ``arroyo.strategies.run_task_with_multiprocessing.batch.output.overflow``
+    3. ``arroyo.strategies.run_task_with_multiprocessing.batch.backpressure``
+
+    If ``batch.input.overflow`` is emitted at all, arroyo ran out of memory for
+    batching and started breaking up your batches into smaller ones. You want
+    to increase ``input_block_size`` in response. Note that when you do this,
```
lynnagara, line 362: There was a conversation in slack about this being fairly harmless. Should we add that here?

fpacifici: It may be one. It depends on the consumer. 

See my comment above. 
What makes tuning this more complex than running an experiment on the Large Hadron Collider is that there are three quantities that interact here, thus you are trying to find the sweet spot on a three dimensional space.
- The message size. Generally the message size has to do with the processing time, outliers can throw everything off.
- input batch size. It can drive how many processes you run at the same time but that is influenced by the message size as well which impacts the processing time for a batch.
- output batch size. It can wreck your throughput if you get it wrong. And it is influenced by the input batch size. If the input batch size increases you have more chances to hit the max memory size in output and get backpressure from the following step. 

So in principle you control two parameters, but the effect of the two parameters is largely impacted by the average and outlier message size -> which drives the processing time.

Now not sure what is the best way to explain this in the doc in a way that could be understood, or whether the architecture has just to change.

lynnagara: Is the way to get rid of it to keep increasing the input blocks though?

In Sentry's indexer I believe this is tackled with the 2 levels of batch sizes specified here. https://github.com/getsentry/sentry/blob/8c734a4c4eccec6fa68d0e193e1d5ea87b4b6045/src/sentry/runner/commands/run.py#L601-L604

But this is a bit complex for the user as there are a lot of parameters to tune - and I worry that we are essentially suggesting that all consumers do it like this.

untitaker: > thus you are trying to find the sweet spot on a three dimensional space.

yes but I believe that it is out of scope for the section on input/output buffers. The entire tuning process should come _after_ getting rid of _all_ overflow IMO, and we should really hammer down that point.

Otherwise you are dealing with three parameters that control input batch size (max time/max size/buffer size), and one parameter that controls output batch size (buffer size). It is just added complexity on top of dealing with three dimensions, and we haven't seen significant cost yet when increasing block size beyond what is needed.

untitaker: > you are trying to find the sweet spot on a three dimensional space.

have we seen any issues with choosing too large values for input and output block size?

if the answer is no, I think we can make this more tractable in practice than a three-dimensional space.

We can tell people that, before tuning, they should overcommit on all block sizes. This gets rid of output batch size as a dimension (it is equal to input batch size), and reduces the number of CLI parameters influencing input batch size.

This is what I was suggesting with the "How to tune your consumer" section at the bottom. I think we don't have to explain all the theory, we just have to come up with a process/runbook to follow. Tell them in which order to tweak those parameters, and which measurements to look at and optimize for.

untitaker: > In Sentry's indexer I believe this is tackled with the 2 levels of batch sizes specified here.

Increasing the input blocks is controlled with `--input-block-size` -- the two batching parameters in Sentry don't control input block size.

I think we should reconsider exposing input and output block size at all. I'd much rather re-allocate on the fly whenever overflow happens, long-term.

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -299,21 +301,104 @@ class RunTaskWithMultiprocessing(
...
+    If those two metrics don't line up, your consumer is not bottlenecked on
+    number of processes. That's a good thing, you want to have some reserve
+    capacity. But it means that increasing ``num_processes`` will not make your
+    consumer faster.
+
+    Batching
+    ~~~~~~~~
+
+    Arroyo sends messages in batches to subprocesses. The cost of a batch is
+    amortized with larger batches. ``max_batch_size`` and ``max_batch_time``
```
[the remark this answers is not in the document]
untitaker: i elaborated a bit on the risks of large batches, but I am curious about this comment

> The issue is that tuning this processor is really hard and making it easy just with docs and training will not actually work.

Do you feel that batching is significantly more difficult with this processor than with other batching mechanisms? Is that because of block sizes or something else?

untitaker:
Since the general topic of metrics docs have come up, here's a POC: https://github.com/getsentry/arroyo/pull/245

lynnagara (approved):

PR #245, opened 2023-05-25
ref: Global metrics registry

untitaker:
Register all metrics in the entire codebase in a Literal, and put
the literal in docs.

This is a bit less overhead than a global enum and requires fewer
changes in strategies that already pass metrics as strings.


lynnagara (approved):
😍

PR #246, opened 2023-05-25
Use structlog for logging

davidtsuk:
This PR transitions logging to use structlog as it has better compatibility with GCP. Currently, this uses a stdlib logging wrapper.

### DIFF
--- a/arroyo/backends/abstract.py
+++ b/arroyo/backends/abstract.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
-import logging
+import structlog
 from abc import ABC, abstractmethod, abstractproperty
 from concurrent.futures import Future
 from typing import Callable, Generic, Mapping, Optional, Sequence, Union
+from arroyo.environment import setup_logging
 
 from arroyo.types import BrokerValue, Partition, Topic, TStrategyPayload
 
-logger = logging.getLogger(__name__)
+setup_logging()
+logger = structlog.get_logger().bind(module=__name__)
 
 
 class Consumer(Generic[TStrategyPayload], ABC):
--- a/arroyo/environment.py
+++ b/arroyo/environment.py
@@ -1,0 +1,32 @@
+import logging
+
+import structlog
+from structlog.types import EventDict
+
+
+def add_severity_attribute(
+    logger: logging.Logger, method_name: str, event_dict: EventDict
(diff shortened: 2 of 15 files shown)


arroyo/backends/abstract.py
```diff
@@ -1,13 +1,15 @@
...
+import structlog
 from abc import ABC, abstractmethod, abstractproperty
 from concurrent.futures import Future
 from typing import Callable, Generic, Mapping, Optional, Sequence, Union
+from arroyo.environment import setup_logging
 
 from arroyo.types import BrokerValue, Partition, Topic, TStrategyPayload
 
-logger = logging.getLogger(__name__)
+setup_logging()
```
evanh, line 11: If this call is always required before calling `structlog.get_logger` then I would wrap it in a single helper function.

```
def get_logger(name):
    setup_logging()
    return structlog.get_logger().bind(module=name)
```

arroyo/environment.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Set the severity attribute for Google Cloud logging ingestion
+    """
+    event_dict["severity"] = event_dict["level"].upper()
+    del event_dict["level"]
+    return event_dict
+
+
+def setup_logging() -> None:
+    structlog.configure(
```
lynnagara, line 19: Shouldn't this get called once, rather than at the top of every module?

arroyo/environment.py
```diff
@@ -0,0 +1,32 @@
...
+        wrapper_class=structlog.stdlib.BoundLogger,
+        logger_factory=structlog.stdlib.LoggerFactory(),
+        processors=[
```
lynnagara, line 22: How does this work with custom processors, such as our Sentry integration? Doesn't this mean that users can't configure the logger?

lynnagara (commented):
Not sure I get the motivation for this.

I know the the `Snuba` service is moving it's internal logging to use structlog, but I'm not sure if it's reasonable to expect all the open source libraries that are used by Snuba to also transition onto structlog.

Since the configure() call now gets made from every module of this library, how would the user configure the logger?

lynnagara:
@davidtsuk Any update on this or can it be closed?

davidtsuk:
I'm gonna close it

PR #249, opened 2023-06-06
feat(metrics): Allow reconfiguring the metrics backend

lynnagara:
There are some rare situations where we might want to reconfigure the metrics backend to something else after a while. This came up in the DLQ consumer code, which can be instructed to consume from multiple queues and apply different policies. In this scenario, we want to tag the metrics very differently each time.

### DIFF
--- a/arroyo/utils/metrics.py
+++ b/arroyo/utils/metrics.py
@@ -1,14 +1,11 @@
 from __future__ import annotations
 
-import logging
 from abc import abstractmethod
 from typing import Any, Mapping, Optional, Protocol, Union, runtime_checkable
 
 from arroyo.utils.metric_defs import MetricName
 
 Tags = Mapping[str, str]
-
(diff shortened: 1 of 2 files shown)


arroyo/utils/metrics.py
```diff
@@ -113,11 +116,12 @@ def decrement(self, value: float = 1.0) -> None:
 
 def configure_metrics(metrics: Metrics) -> None:
```
untitaker, line 117: can we add a function argument here that allows you to opt into overriding? In sentry we have caught quite a few initialization bugs by arroyo raising because metrics were configured twice.

```
def configure_metrics(metrics: Metircs, override=False): ...
```

untitaker (changes requested):

arroyo/utils/metrics.py
lynnagara: Good idea, updated

PR #253, opened 2023-06-13
feat: Avoid spinning on MessageRejected

lynnagara:
This was done in https://github.com/getsentry/sentry/blob/43de083a4b874ab23de5db2c8f7f2958a3cb70ab/src/sentry/processing/backpressure/arroyo.py#L40, should probably be an Arroyo default


untitaker (approved):
you know, i was adamant that we punt on this issue and later solve it "properly" somehow in arroyo, but this is probably good enough. and honestly i hadn't considered it that we just merge the same hack from consumers into arroyo

PR #254, opened 2023-06-15
perf: Don't unnecessarily recreate strategy in revocation callback

lynnagara:
It's very important for the revocation callback to finish quickly as it blocks assignments on the entire consumer group and causes slower than necessary rebalancing. Let's not unnecessarily create the strategy here, it will be done in on_assign.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -225,24 +225,6 @@ class StreamProcessor(Generic[TStrategyPayload]):
             if partitions:
                 _close_strategy()
 
-                # Recreate the strategy if the consumer still has other partitions
-                # assigned and is not closed or errored
-                try:
-                    current_partitions = self.__consumer.tell()
-                    if len(current_partitions.keys() - set(partitions)):
-                        active_partitions = {
-                            partition: offset
-                            for partition, offset in current_partitions.items()
-                            if partition not in partitions
-                        }
-                        logger.info(
-                            "Recreating strategy since there are still active partitions: %r",
-                            active_partitions,
-                        )
-                        _create_strategy(active_partitions)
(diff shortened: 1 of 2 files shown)


ayirr7:
This change makes sense to me, thanks

ayirr7 (approved):

nikhars:
Not sure how I feel about this. Is the following sequence of events possible

1. Assign partition 1
2. Assign partition 2
3. Revoke partition 1
4. Assign partition 3
5. Revoke partition 3
6. Rebalancing over

In this scenario the pod is still assigned partition 2 and should consume from it. But with the change it would not.

lynnagara:
That scenario isn't possible in arroyo since incremental rebalancing isn't supported

lynnagara:
I realised this probably doesn't help much though since we never actually recreate the strategy in practice, so may not be worth making this change

untitaker:
> In this scenario the pod is still assigned partition 2 and should consume from it. But with the change it would not.

I don't understand how that is true. We are still closing the strategy, it would be re-created on next poll.

untitaker:
How about we lazily create the partitions whenever we have a message to process? Then we never have to recreate a strategy in any assign/revoke callback, we just have to close it.

lynnagara:
> How about we lazily create the partitions whenever we have a message to process? Then we never have to recreate a strategy in any assign/revoke callback, we just have to close it. Meaning that consecutive partition assignments and revocations don't create strategies repeatedly.

How would we go about passing the assigned partition list to the strategy? This is currently required by `Strategy.create_with_partitions()`

lynnagara:
@untitaker attempt to reuse strategies: https://github.com/getsentry/arroyo/pull/255

PR #255, opened 2023-06-16
feat: Avoid recreating strategies

lynnagara:
Based on comment from @untitaker in https://github.com/getsentry/arroyo/pull/254#issuecomment-1593185164. Throwing this up as draft as it's not yet a properly fleshed out idea.

Motivation is that we need the on_revoke and on_assign callback to be as fast as possible, and most of the time the strategies can be reused as-is.


arroyo/processing/processor.py
```diff
@@ -189,18 +189,20 @@ def _close_strategy() -> None:
                 "%r exited successfully, releasing assignment.",
                 self.__processing_strategy,
             )
-            self.__processing_strategy = None
+            self.__processing_strategy.on_assignment_update({}, self.__join_timeout)
```
untitaker, line 192: we still close + flush here, otherwise yeah that's the rough idea

PR #259, opened 2023-06-20
fix: Commit offsets even when topic is empty

untitaker:
Currently it's not possible for arroyo to ever hit consumer lag=zero if
nobody is constantly producing to the topic.

### DIFF
--- a/arroyo/processing/strategies/commit.py
+++ b/arroyo/processing/strategies/commit.py
@@ -17,7 +17,7 @@ class CommitOffsets(ProcessingStrategy[Any]):
         self.__commit = commit
 
     def poll(self) -> None:
-        pass
+        self.__commit({})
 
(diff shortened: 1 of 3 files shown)


lynnagara (approved):

untitaker:
for context, this bug matters only in the qe env, where we:

1. fill a topic with test data
2. reset offset to earliest
3. run consumer
4. poll consumer lag from statsd, and terminate everything once it hits 0

it never hit zero, so all the tests are just hanging for ingest-v2 consumers right now

unclear yet why it does not affect metrics-indexer and replays, which also have a qe env setup

lynnagara:
In the indexer, the preceding produce step calls next_step.submit() with the relevant offsets in poll upon checking for produced message status.

PR #261, opened 2023-06-23
perf: Factory with cached strategy

lynnagara:
Alternative to https://github.com/getsentry/arroyo/pull/255 that doesn't involve changing Arroyo's API so much.


untitaker (commented):
creating a strategy isn't expensive, join() ing is. I don't think this helps because processor is still calling join() and waits for flushing

PR #262, opened 2023-06-26
feat: Add liveness healthcheck

### LINKED ISSUE #247: Make arroyo serve a liveness/readiness checks for k8s
This idea was provided by @beezz. Arroyo consumers can die because of many reasons. The largest category is about the main thread being blocked. Main thread can be blocked due to bugs during rebalancing (as observed in #217 #237), or because the user-defined strategy (or RunTask) misbehaves.

When we fail to poll, Kafka already kicks us out of the broker after `max.poll.interval`. But the pod is still there, doing nothing.

The suggestion by michal is as follows: We should, everytime we poll, touch a file (debounced). If the file has not been recently touched, k8s notices and deletes the pod.

untitaker:
fix #247

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -148,7 +148,7 @@ class StreamProcessor(Generic[TStrategyPayload]):
         self.__commit_policy_state = commit_policy.get_state_machine()
         self.__join_timeout = join_timeout
         self.__healthcheck_file = healthcheck_file
-        self.__healthcheck_file_touched_at = None
+        self.__healthcheck_file_touched_at: Optional[float] = None
         self.__shutdown_requested = False
 
         # Buffers messages for DLQ. Messages are added when they are submitted for processing and
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -34,7 +34,6 @@ from arroyo.utils.metrics import get_metrics
 logger = logging.getLogger(__name__)
 
 METRICS_FREQUENCY_SEC = 1.0  # In seconds
-HEALTHCHECK_MAX_FREQUENCY_SEC = 1.0  # In seconds
 
 F = TypeVar("F", bound=Callable[[Any], Any])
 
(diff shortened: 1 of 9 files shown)


lynnagara (approved):

docs/source/healthcheck.rst
```diff
@@ -0,0 +1,60 @@
...
+Kafka has a setting called ``max.poll.interval.ms`` for this that tells Kafka
+to kick the consumer out of the broker after this many milliseconds of not polling.
+
+You can pass this option into :py:class:`arroyo.backends.kafka.consumer.KafkaConsumer` like so:
+
+.. code-block:: Python
+
+   consumer = KafkaConsumer(
+       {
+           "max.poll.interval.ms": 5 * 60, # default 5 minutes
```
mwarkentin, line 16: This looks like 300ms based on the unit, or it's missing a `* 1000`

untitaker, line 16:
```suggestion
           "max.poll.interval.ms": 300000, # default 5 minutes
```

docs/source/healthcheck.rst
```diff
@@ -0,0 +1,60 @@
...
+        image: registry.k8s.io/busybox
+        args:
+        - /bin/sh
+        - -c
+        - rm /tmp/health.txt
+        livenessProbe:
+          exec:
+            command:
+            - cat
+            - /tmp/healthy
```
dbanda, line 58: This k8s example doesn't align with the code. I'm guessing you copy pasted it. This might be a dump question, but in the case of arroyo, would the liveness command would have to be a unix command to checks the mtime of the file?

lynnagara:
Can we add this as a strategy instead of building it inside the stream processor? I think it would be better to avoid making that file more complex for this feature if we can avoid it.

docs/source/healthcheck.rst
untitaker: Yeah that's a copypaste error. I accidentally replaced the container entrypoint with the healthcheck command. fixed

I could check mtime, instead I am deleting the file every n seconds. If it stops being recreated, the consumer is dead

beezz (approved):
🚀

untitaker:
@lynnagara sure, tbh i find it harder to reason about as separate strategy since calling `poll()` regularly is not really part of the strategy contract.. left the test e2e as is

arroyo/processing/processor.py
```diff
@@ -145,6 +146,8 @@ def __init__(
 
         self.__commit_policy_state = commit_policy.get_state_machine()
         self.__join_timeout = join_timeout
+        self.__healthcheck_file = healthcheck_file
+        self.__healthcheck_file_touched_at: Optional[float] = None
```
lynnagara, line 150: Can you revert the changes on this file?

untitaker: whoops

tests/processing/test_processor.py
```diff
@@ -545,3 +548,40 @@ def test_dlq() -> None:
     processor._run_once()
 
     assert dlq_policy.producer.produce.call_count == 1
+
+
+def test_healthcheck(tmpdir: py.path.local) -> None:
```
lynnagara, line 553: Can we move this test out of the processor too?

untitaker: I want to test processor + healthcheck e2e. testing in isolation makes no sense. where do you want me to put the test?

lynnagara: Ah I see. The rest of the strategies are only tested in isolation so I was thinking under tests/processing/strategies with the rest. This is fine though.

lynnagara (approved):

Issue #257, opened 2023-06-16
Filter step should emit metrics about dropped events


lynnagara:
Looked into replacing some hand rolled strategies with Filter and other built in Arroyo strategies, and noticed some metrics are missing that are probably useful to end users.

e.g. https://github.com/getsentry/snuba/blob/2705677b817be29abd83b69bedd7bcb379df2234/snuba/subscriptions/executor_consumer.py#L357

The filter step could just automatically record this.

ayirr7:
One question, the linked code obtains the entity name by decoding a scheduled subscription task, and then emits the `skipped_execution` metric with this entity name. Do we want to retain this information if emitting this metric in the FilterStep in Arroyo? I'm not sure how we could obtain it.

ayirr7:
@lynnagara One question, the linked code obtains the entity name by decoding a scheduled subscription task, and then emits the `skipped_execution` metric with this entity name. Do we want to retain this information if emitting this metric in the FilterStep in Arroyo? I'm not sure how we could obtain it.

lynnagara:
No need to explicitly include any tags in here as the metric interface allows the user to configure any custom tags they like when it's initialized. We already attach the storage, consumer group and other information to all arroyo metrics emitted from Snuba.

lynnagara:
Done in https://github.com/getsentry/arroyo/pull/265

PR #265, opened 2023-06-27
metrics: Add metrics about dropped messages in FilterStep

ayirr7:
For https://github.com/getsentry/arroyo/issues/257 

It seemed to me that we should emit this metric in both scenarios, whether a CommitPolicy was passed in or not. Might have missed something though.

Also, if there are other metrics that would be helpful to add in `FilterStep` strategy, I can include them in this PR.

### DIFF
--- a/arroyo/processing/strategies/filter.py
+++ b/arroyo/processing/strategies/filter.py
@@ -82,15 +82,10 @@ class FilterStep(ProcessingStrategy[Union[FilteredPayload, TStrategyPayload]]):
             for partition in message.committable:
                 self.__uncommitted_offsets.pop(partition, None)
             self.__next_step.submit(message)
-        elif self.__commit_policy_state is not None:
-            # first scenario, where a CommitPolicy is passed in
-            # and messages are dropped
-            self.__uncommitted_offsets.update(message.committable)
-            self.__metrics.increment("arroyo.strategies.filter_step.dropped_messages")
-        elif self.__commit_policy_state is None:
-            # second scenario, where a CommitPolicy is not passed in
-            # and messages are dropped
-            self.__metrics.increment("arroyo.strategies.filter_step.dropped_messages")
+        else:
+            self.__metrics.increment("arroyo.strategies.filter.dropped_messages")
+            if self.__commit_policy_state is not None:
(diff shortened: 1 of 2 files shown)


arroyo/processing/strategies/filter.py
```diff
@@ -80,7 +83,14 @@ def submit(
...
             self.__next_step.submit(message)
         elif self.__commit_policy_state is not None:
+            # first scenario, where a CommitPolicy is passed in
+            # and messages are dropped
             self.__uncommitted_offsets.update(message.committable)
+            self.__metrics.increment("arroyo.strategies.filter_step.dropped_messages")
+        elif self.__commit_policy_state is None:
+            # second scenario, where a CommitPolicy is not passed in
+            # and messages are dropped
+            self.__metrics.increment("arroyo.strategies.filter_step.dropped_messages")
```
lynnagara, line 93: nit: The 2 elif statements make it harder to parse
```suggestion
        else:
            self.__metrics.increment("arroyo.strategies.filter_step.dropped_messages")
            if self.__commit_policy_state is not None:
                self.__uncommitted_offsets.update(message.committable)
```

arroyo/processing/strategies/filter.py
```diff
@@ -80,7 +83,14 @@ def submit(
                 self.__uncommitted_offsets.pop(partition, None)
             self.__next_step.submit(message)
         elif self.__commit_policy_state is not None:
+            # first scenario, where a CommitPolicy is passed in
+            # and messages are dropped
             self.__uncommitted_offsets.update(message.committable)
+            self.__metrics.increment("arroyo.strategies.filter_step.dropped_messages")
```
lynnagara, line 89: Can we make this:
```suggestion
            self.__metrics.increment("arroyo.strategies.filter.dropped_messages")
```
It kinda bugs me that a couple of the steps are suffixed with "step" and I think we should standardise it.

ayirr7: Done

arroyo/processing/strategies/filter.py
ayirr7: Yep, makes sense

lynnagara (approved):
Thanks!

PR #270, opened 2023-06-29
feat: Automatically resize blocks if they get too small

untitaker:
This is a naive implementation that resizes input/output blocks if they
become too small.

Blocks are reallocated once the overflowing batch is successfully
drained.

This will make startup of a consumer slower, but over time hopefully all
blocks eventually reach a state where they won't overflow anymore.

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -309,6 +309,21 @@ class RunTaskWithMultiprocessing(
         ``arroyo.strategies.run_task_with_multiprocessing.batch.output.overflow``
         metric is incremented.
 
+    :param resize_input_blocks: Experimental feature, whether input blocks
+        should be dynamically resized if they end up being too small for batches.
+        This can technically cause unbounded memory consumption and it is
+        recommended to also configure `max_input_block_size` if this option is
+        used.
+
+    :param resize_output_blocks: Experimental feature, same as
+        `resize_input_blocks` but for the output blocks.
+
+    :param max_input_block_size: If automatic resizing is enabled, this sets an
+        upper limit on how large those blocks can get.
+
+    :param max_output_block_size: Same as `max_input_block_size` but for output
+        blocks.
+
     :param initializer: A function to run at the beginning of each subprocess.
 
         Subprocesses are spawned without any of the state of the parent
@@ -413,12 +428,21 @@ class RunTaskWithMultiprocessing(
         max_batch_time: float,
         input_block_size: int,
         output_block_size: int,
+        resize_input_blocks: bool = False,
+        resize_output_blocks: bool = False,
+        max_input_block_size: Optional[int] = None,
+        max_output_block_size: Optional[int] = None,
         initializer: Optional[Callable[[], None]] = None,
     ) -> None:
         self.__transform_function = function
         self.__next_step = next_step
         self.__max_batch_size = max_batch_size
         self.__max_batch_time = max_batch_time
+
+        self.__resize_input_blocks = resize_input_blocks
+        self.__resize_output_blocks = resize_output_blocks
+        self.__max_input_block_size = max_input_block_size
+        self.__max_output_block_size = max_output_block_size
 
         self.__shared_memory_manager = SharedMemoryManager()
         self.__shared_memory_manager.start()
@@ -454,6 +478,8 @@ class RunTaskWithMultiprocessing(
                     ]
                 ],
                 AsyncResult[ParallelRunTaskResult[TResult]],
+                bool,  # was the input block too small?
+                bool,  # was the output block too small?
             ]
         ] = deque()
         self.__invalid_messages = InvalidMessageState()
@@ -481,7 +507,7 @@ class RunTaskWithMultiprocessing(
 
         signal.signal(signal.SIGCHLD, handle_sigchld)
 
-    def __submit_batch(self) -> None:
+    def __submit_batch(self, input_block_too_small: bool) -> None:
         assert self.__batch_builder is not None
         batch = self.__batch_builder.build()
         logger.debug("Submitting %r to %r...", batch, self.__pool)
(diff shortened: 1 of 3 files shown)


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -413,13 +428,22 @@ def __init__(
         max_batch_time: float,
         input_block_size: int,
         output_block_size: int,
+        resize_input_blocks: bool = False,
+        resize_output_blocks: bool = False,
+        max_input_block_size: Optional[int] = None,
+        max_output_block_size: Optional[int] = None,
```
lynnagara, line 434: With so many extra parameters, this seems like it would be more complicated rather than less to figure out what to pass here

untitaker: yeah I just want to make this opt-in for now. I think long-term we should give  `input_block_size` and default, and enable auto-resizing by default. Then we still have a lot of parameters, but people don't _need_ to set any of them.

Right now they _need_ to configure `input_block_size`, and we can't give them a default.

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -604,12 +646,54 @@ def __check_for_results_impl(self, timeout: Optional[float] = None) -> None:
...
+            and (
+                self.__max_input_block_size is None
+                or self.__max_input_block_size > old_input_block.size * 2
+            )
+        ):
+            self.__metrics.increment(
+                "arroyo.strategies.run_task_with_multiprocessing.batch.input.resize"
+            )
+            new_input_block = self.__shared_memory_manager.SharedMemory(
+                old_input_block.size * 2
```
lynnagara, line 668: This seems like it could trigger quite a few resizes depending on the values passed. Is there another way to calculate the sharedmemory size based on prior seen message sizes?

arroyo/processing/strategies/run_task_with_multiprocessing.py
lynnagara: If the user knows what the max is supposed to be, what's the rationale for them to start with a smaller number at all, rather than just passing the max into `input_block_size` and `output_block_size`

arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -604,12 +646,54 @@ def __check_for_results_impl(self, timeout: Optional[float] = None) -> None:
...
+            and self.__resize_input_blocks
+            and (
+                self.__max_input_block_size is None
+                or self.__max_input_block_size > old_input_block.size * 2
+            )
+        ):
+            self.__metrics.increment(
+                "arroyo.strategies.run_task_with_multiprocessing.batch.input.resize"
+            )
+            new_input_block = self.__shared_memory_manager.SharedMemory(
```
fpacifici, line 667: Knowing that memory settings for Kubernetes are static (we cannot resize the memory assigned to a pod) and that you cannot exceed the memory allocated (OOMKill), do we have a reason to ever resize our input/output blocks instead of just taking all the available memory at the start ? It seems it would be much easier to just statically create memory blocks that used all the available memory of the pod and not try to change anything.

untitaker: > do we have a reason to ever resize our input/output blocks instead of just taking all the available memory at the start

there's still a ton of stuff outside of arroyo that takes memory, but even if not I don't think it's a good idea to consume significantly more memory than needed. When we deploy a new service we configure k8s limits based on projected memory usage, and later optionally adjust based on average memory usage. If arroyo defaults to consuming all pod memory, we lose insight into how much memory we actually need and how much cost we could save.

fpacifici: > but even if not I don't think it's a good idea to consume significantly more memory than needed

Why?  That memory is already allocated and not usable by others anyway. 
If you are concerned of not having visibility on the actual usage, why not having specific metrics for that? It seems an easier, safer system with fewer moving parts and fewer failure modes.

untitaker: The goal is to eliminate tuning parameters the user has to tweak to get optimal consumer performance. If we manage to do that, then we can start thinking of removing some of those options (as they are not required and have optimal defaults) and therefore moving parts in that sense.

An engineer of the product team should not have to think about how much memory their pod is going to consume and tune Arroyo parameters based off of it.

I don't think this is possible at all with a static approach, because it requires the author of the consumer to think about how much memory their pod has (unclear, gets adjusted by ops), how much their regular code consumes per-process (entirely unclear, especially in a shared codebase like sentry where tons of random stuff gets imported at every CLI invocation) and then think about how much of the remainder can be allocated to input/output blocks.

If you are suggesting a static approach that is also zero-config, I don't know how that would work. Does it mean that arroyo determines free memory and allocates it evenly divided for input/output blocks? And is it evenly, or do input blocks get more than output blocks? And what does it do on a dev machine where there's no k8s request/limit per-consumer?

untitaker: > If you are concerned of not having visibility on the actual usage, why not having specific metrics for that?

I think this is possible but it feels like Java/node heap tuning parameters and I would like to avoid that sort of experience as well.

arroyo/processing/strategies/run_task_with_multiprocessing.py
untitaker: I think that's a good idea, we can keep track of the batch size and use it as input for reallocating. I'll have to see how to do that.. I don't think I can use our existing batch size metrics because they don't emit the right value when batches are split up due to input/output overflow.

untitaker:
after talking to lyn, we decided that it makes more sense to expose `input_block_size=None` (as default) instead of adding additional flags. The initial block sizes are hardcoded at 16kB in those cases.

There was also an idea to use the message size to guide by how much the input block should be reallocated. I tried this and found it was quite hard to implement, I don't think it will significantly impact startup performance.

lynnagara (approved):

PR #273, opened 2023-07-14
perf: Add a small sleep if backpressure applied to multiprocessing step

lynnagara:
This check runs in a tight loop if there is backpressure from downstream step, let's add a small sleep to avoid spinning here

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -527,6 +527,7 @@ class RunTaskWithMultiprocessing(
                 )
             except NextStepTimeoutError:
                 if deadline is None or deadline > time.time():
+                    time.sleep(0.01)
                     continue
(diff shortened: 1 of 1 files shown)


evanh (approved):

untitaker (commented):
@lynnagara can you explain why we'd be spinning here? it's not clear to me

lynnagara:
I think it's possible since this line runs in a while loop in join. If we get message rejected from the next step, we just keep submitting over and over again.

untitaker:
if there is backpressure from downstream step i'd expect that `next_step.poll()` makes some form of progress (while blocking for multiple milliseconds), which we call in `__check_for_results_impl`. If it doesn't make progress, `next_step` is just completely stuck and the busy-spinning is a symptom, not a problem.

lynnagara:
Hmm, you're right. I missed the call to poll() in there, closing

PR #276, opened 2023-08-10
fix(run_task_in_threads): Commit offsets even when consumer is idle

untitaker:
When using run_task_in_threads, the last batch of commits might be
dropped because poll() is not forwarded to the commit strategy.

This presents with the same exact symptoms as #259, but is only present
in consumers that use RunTaskInThreads.


onewland (approved):
this looks fine to me but I'm low confidence on what conditions one may want/not want to move poll() to the next step

untitaker:
I think there may be more scenarios in one which might want to poll, but I am inclined to discover them as they cause problems in prod. Overly eager polling might cause performance problems (such as polling next-step twice in succession)

PR #277, opened 2023-08-15
fix(multiprocessing): Make close() call the next steps close()

nikhars:
We are missing the call to next step's close() call in multiprocessing. In cases where the next step is Produce the consumer just hangs because the shutdown flag is not set.

### DIFF
--- a/arroyo/processing/strategies/produce.py
+++ b/arroyo/processing/strategies/produce.py
@@ -124,5 +124,6 @@ class Produce(ProcessingStrategy[Union[FilteredPayload, TStrategyPayload]]):
             self.__next_step.poll()
             self.__next_step.submit(message)
 
+        self.__producer.close()
         self.__next_step.close()
         self.__next_step.join(remaining)
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -669,6 +669,8 @@ class RunTaskWithMultiprocessing(
         if self.__batch_builder is not None and len(self.__batch_builder) > 0:
             self.__submit_batch()
 
+        self.__next_step.close()
+
     def terminate(self) -> None:
         self.__closed = True
 
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -669,8 +669,6 @@ class RunTaskWithMultiprocessing(
         if self.__batch_builder is not None and len(self.__batch_builder) > 0:
             self.__submit_batch()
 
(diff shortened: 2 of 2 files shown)


nikhars:
@untitaker @lynnagara - Is calling next step's close() method incorrect here? The tests seem to be failing which is a bit surprising. Before I start digging into the test failures I would like to know whether what the PR is proposing is correct or not. For context: @phacops [PR](https://github.com/getsentry/sentry/pull/54154) for ingest spans consumer test gets hung right now. Specifically this test https://github.com/getsentry/sentry/pull/54154/files#diff-0626054a3c2af5ad276c422567321f4bc326b8b091ff13e652c34f4560396e79R14. I narrowed it down to the KafkaProducer being stuck in [this](https://github.com/getsentry/arroyo/blob/main/arroyo/backends/kafka/consumer.py#L652-L653) while loop forever because it never gets called with close(). The strategy for reference is [here](https://github.com/getsentry/sentry/pull/54154/files#diff-973384a59a8a97829c8a4ceb385fc4fa74a3ac9b250191da9e53ef22f95a518eR130-R142)

phacops:
I tried this fix but it doesn't seem to fix the issue unfortunately, still stuck in this loop.

lynnagara:
next_step.close() is already being called in join. We can't call it here because then we wouldn't be able to keep forwarding remaining messages to the next step during shutdown

nikhars:
I found the problem. The `Produce` strategies join() does not call the producer's close(). Since the producer is running in a different thread and it's close method is not called, the thread hangs which causes the program to not terminate.

@phacops - Can you test this fix with your PR?

untitaker (approved):
please rename the pr and ideally add test

lynnagara (changes requested):
I don't think this will work. Since the producer is passed into the strategy, it will be closed on rebalance and will cease to work later. The producer can outlive the strategy - we don't need to close it every time we close the strategy due to rebalance.

Since the producer is created by the user of Arroyo and passed to the strategy, it's up to the user to close the producer in their code. (For example, in Snuba we do this like https://github.com/getsentry/snuba/blob/793c43889d89776f5d5ecc2268b861384b3d7e6a/snuba/cli/subscriptions_scheduler.py#L207 or https://github.com/getsentry/snuba/blob/793c43889d89776f5d5ecc2268b861384b3d7e6a/snuba/consumers/consumer_builder.py#L313-L314)

In general, I think Arroyo should only manage closing resources it creates internally and not those that are passed in.

phacops:
The fix in this PR works for the issue I'm facing.

@lynnagara I'd love if I didn't have to pass a producer to the `Produce` step in that case. As a user, I don't really want to be responsible but a `Producer`that I then have to close "properly".

PR #281, opened 2023-08-18
fix(produce): Apply backpressure instead of crashing

lynnagara:
If we get local queue full, let's raise MessageRejected to slow down the consumer.


volokluev (commented):
Can we add a test for this?

untitaker (approved):

PR #283, opened 2023-08-29
fix: Ensure carried over message is in buffer

lynnagara:
Since `_run_once` can early return (https://github.com/getsentry/arroyo/blob/6286c7921978065edeb5ce72d829031f952d2e5e/arroyo/processing/processor.py#L369) it was possible that a message was never placed in `self.buffered_messages`. If we try to retreive it later, it can crash the consumer.

This is suspected to be the cause of the `Invalid message not found in buffer` messages we saw in prod.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -377,8 +377,7 @@ class StreamProcessor(Generic[TStrategyPayload]):
                     message = (
                         Message(self.__message) if self.__message is not None else None
                     )
-                    if not message_carried_over:
-                        self.__buffered_messages.append(self.__message)
+                    self.__buffered_messages.append(self.__message)
                     self.__processing_strategy.submit(message)
 
                     self.__metrics_buffer.incr_timing(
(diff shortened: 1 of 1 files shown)


untitaker (approved):
I believe this data flow would be more comprehensible if we write into `self.__buffered_messages` at the same time we set `self.__messages` (i.e. during rdkafka poll)

lynnagara:
Wondering if it's better to remove the early return on line 369. This way we might be putting the same message in the buffer multiple times if we have backpressure.

lynnagara:
@untitaker you're right, does this look right to you now?

lynnagara:
> i believe we should ensure that `poll` returns successfully at least once before we send a message in. otherwise I suspect we might discover new ways to overload processing strategies.

do you mean `strategy.poll()` rather than `consumer.poll()`? I don't think this can overload the strategy - we never poll on the consumer again or run this block of code for as long as there is a carried over message.

untitaker:
>      do you mean strategy.poll() rather than consumer.poll()? 

yes, I mean strategy poll

untitaker (approved):

PR #285, opened 2023-09-13
feat: Add supported configuration

lynnagara:
Add max.poll.interval.ms and session.timeout.ms. These allows for easier configuration of these settings.

Honestly not sure if this is even useful and we should drop this "validation" instead


lynnagara:
On second thoughts, kinda want to delete this all

PR #289, opened 2023-09-27
ref: A better commit log format

lynnagara:
I really don't want to port this abomination to Rust, let's simplify the commit log format

Two main changes:
- uses float instead of weird datetime format we had before
- avoids putting important information in headers. This was a hack to avoid changing the message format earlier, but it should be avoided. The timestamp is now in the payload of the message

This is the first of 3 changes, encode method is unchanged, the decode method supports the new format.

### DIFF
--- a/arroyo/backends/kafka/commit.py
+++ b/arroyo/backends/kafka/commit.py
@@ -1,3 +1,4 @@
+import json
 from datetime import datetime
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


evanh (approved):

untitaker (approved):

lynnagara:
On second thoughts, gonna move the encode change into a separate PR for deploy safety reasons

PR #290, opened 2023-09-28
ref: A better commit log format pt2

lynnagara:
Follow up to https://github.com/getsentry/arroyo/pull/289

### DIFF
--- a/arroyo/backends/kafka/commit.py
+++ b/arroyo/backends/kafka/commit.py
@@ -1,3 +1,4 @@
+import json
 from datetime import datetime
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


untitaker (approved):

arroyo/backends/kafka/commit.py
```diff
@@ -6,29 +7,31 @@
...
 
 
 class CommitCodec(Codec[KafkaPayload, Commit]):
     def encode(self, value: Commit) -> KafkaPayload:
         assert value.orig_message_ts is not None
 
+        payload = json.dumps(
+            {
+                "offset": value.offset,
+                "orig_message_ts": datetime.timestamp(value.orig_message_ts),
```
fpacifici, line 22: Is this for the alerts e2e SLO ?

lynnagara: Not really, this is the same information that was already there, just in the body not the header now so we don't need to do the custom decoding dance

PR #293, opened 2023-10-04
ref: Timestamp is not optional on commit log

lynnagara:
It was only this way for rollout, it's now there on all messages for years

### DIFF
--- a/arroyo/backends/kafka/commit.py
+++ b/arroyo/backends/kafka/commit.py
@@ -1,6 +1,5 @@
 import json
 from datetime import datetime
-from typing import Optional
 
 from arroyo.backends.kafka import KafkaPayload
 from arroyo.commit import Commit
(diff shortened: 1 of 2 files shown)


arroyo/backends/kafka/commit.py
```diff
@@ -41,13 +40,9 @@ def decode_legacy(self, value: KafkaPayload) -> Commit:
...
-        try:
-            orig_message_ts: Optional[datetime] = datetime.strptime(
-                headers["orig_message_ts"].decode("utf-8"), DATETIME_FORMAT
-            )
-        except KeyError:
-            orig_message_ts = None
-
+        orig_message_ts = datetime.strptime(
+            headers["orig_message_ts"].decode("utf-8"), DATETIME_FORMAT
+        )
```
lynnagara, line 45: This code (decode_legacy) will eventually all be deleted anyway - I'm waiting for the next self hosted release to make the migration smoother

evanh (approved):

PR #296, opened 2023-10-24
perf: Avoid unnecessarily clearing the rdkafka buffer on backpressure

lynnagara:
This is the same fix we did in the Rust version of Arroyo https://github.com/getsentry/snuba/pull/4774

We saw a lot of performance improvements after this change so I'm porting it to Python too.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -142,9 +142,9 @@ class StreamProcessor(Generic[TStrategyPayload]):
 
         self.__message: Optional[BrokerValue[TStrategyPayload]] = None
 
+        # The timestamp when backpressure state started
         self.__backpressure_timestamp: Optional[float] = None
-        # If the consumer is in the paused state, this is when the last call to
-        # ``pause`` occurred.
+        # Consumer is paused after it is in backpressure state for > 1 second
         self.__is_paused = False
 
         self.__commit_policy_state = commit_policy.get_state_machine()
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -350,7 +350,7 @@ class StreamProcessor(Generic[TStrategyPayload]):
                 raise InvalidStateError(
                     "received message when consumer was expected to be paused"
                 )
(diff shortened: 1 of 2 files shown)


arroyo/processing/processor.py
```diff
@@ -142,9 +142,10 @@ def __init__(
 
         self.__message: Optional[BrokerValue[TStrategyPayload]] = None
 
+        self.__backpressure_timestamp: Optional[float] = None
         # If the consumer is in the paused state, this is when the last call to
-        # ``pause`` occurred or the time the pause metric was last recorded.
-        self.__paused_timestamp: Optional[float] = None
+        # ``pause`` occurred.
```
nikhars, line 147: Nit: This comment does not align with how self.__is_paused is used now. Can you update the comment?

arroyo/processing/processor.py
```diff
@@ -394,39 +392,44 @@ def _run_once(self) -> None:
...
 
-                        self.__paused_timestamp = time.time()
                     else:
-                        current_time = time.time()
                         time.sleep(0.01)
-                        if self.__paused_timestamp:
-                            self.__metrics_buffer.incr_timing(
-                                "arroyo.consumer.paused.time",
-                                current_time - self.__paused_timestamp,
-                            )
```
nikhars, line 412: Is there a way to still capture the consumer pause time? If we lose visibility into consumer pauses, troubleshooting consumer bottlenecks becomes hard.

arroyo/processing/processor.py
lynnagara: Fixed

arroyo/processing/processor.py
lynnagara: Oh, we're actually still recording the paused time with the same metric if you look below. I just removed this contraption that re-records on every iteration of the loop and then updates the timestamp because i think that's overkill and makes the code more complicated. Now we just record it once when we unpause, which is what we ended up doing in the Rust code too.

nikhars (approved):

PR #299, opened 2023-11-06
fix: Cleanup backpressure state between assignments

lynnagara:
This is a fix for inc-552

It caused a bug where a message could be submitted to the processing strategy twice immediately after assignment if there was backpressure before the rebalance.


john-z-yang (approved):
This looks good to me. Let's inform the oncall when we bump the version in sentry just in case we missed something and the error shows up again.

PR #301, opened 2023-11-07
fix(dlq): RunTaskWithMultiprocessing supports forwarding downstream invalid message

lynnagara:
Invalid messages raised from strategies downstream of RunTaskWithMultiprocessing are not correctly handled currently. Currently messages from that batch will be re-submitted to the next step multiple times. This change ensures they are correctly re-raised and avoids the duplicate messages downstream.

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -788,7 +788,7 @@ class RunTaskWithMultiprocessing(
                 )
                 break
             except InvalidMessage:
-                pass
+                raise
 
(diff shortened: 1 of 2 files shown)


nikhars (approved):

john-z-yang (approved):
Tested locally with the metrics indexer.

Created 16 partition on the metrics topic and ran 2 indexers, each with 4 processes and prod batch size/timing.

Sent 1560 messages, 1440 of them are valid and 120 of them invalid, and verified that the valid ones all showed up in clickhouse, and invalid ones are in the dlq.

LGTM, thank you so much

PR #302, opened 2023-11-08
ref: Add extra logging for debugging commits

lynnagara:
There is a mystery right now which is causing (probably) very old commit messages to still be present. This seems to be the case even after we switched the cleanup policy on the topic to `compact,delete` from `compact` in all environments so that messages eventually expire (they previously never expired). This adds some logging to help us figure out where/when these old messages came from and helps us determine when the legacy decoder can be safely removed.


volokluev (approved):
🕵️ love a good mystery

PR #303, opened 2023-11-15
ref: Rename backpressure metric

lynnagara:
Rename `arroyo.consumer.paused.time` to `arroyo.consumer.backpressure.time`

Two main motivations for this:
- The new metric matches the name emitted by the Rust consumer, we should align these exactly to make the transition smoother
- The consumer is no longer actually paused immediately when backpressure happens (we now wait until we are in the backpressure state for at least a second as pausing is expensive). So the old metric name is kind of misleading now.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -66,7 +66,7 @@ class InvalidStateError(RuntimeError):
 ConsumerTiming = Literal[
     "arroyo.consumer.poll.time",
     "arroyo.consumer.processing.time",
-    "arroyo.consumer.paused.time",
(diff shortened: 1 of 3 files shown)


untitaker (approved):
i feel that the renaming breaks too many dashboards, and that we shouldn't do it too often with metrics in general. but for this one it's probably fine.

lynnagara:
> i feel that the renaming breaks too many dashboards, and that we shouldn't do it too often with metrics in general. but for this one it's probably fine.

I'll make sure to update (afaik there's one for sentry and one for snuba). I don't think there's any others that need to be changed.

PR #304, opened 2023-11-17
fix: Try fixing InvalidStateError

lynnagara:
Not confident this is going to do anything but I don't really have other ideas. For some reason the consumer occasionally does not seem to be truly paused even when the self.__is_paused flag is set.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -416,6 +416,13 @@ class StreamProcessor(Generic[TStrategyPayload]):
                         self.__consumer.pause([*self.__consumer.tell().keys()])
                         self.__is_paused = True
 
+                        # XXX: This is kind of an experiment. We occasionally see the
+                        # `received message when consumer was expected to be paused` error
+                        # in prod. My only theory for why this happens right now is that pausing
+                        # consumer is somehow not synchronous and we are trying to poll before it is
+                        # truly paused. Add a small sleep to see if it makes any difference.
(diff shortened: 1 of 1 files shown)


untitaker:
my current suspicion is that we are applying backpressure while new partitions are being assigned, somehow. so we pause the consumer here:

https://github.com/getsentry/arroyo/blob/ba85564a91f95b0188d626dd3c03bb855f15726c/arroyo/processing/processor.py#L402

but at the next poll, we get a partition assigned, which is not paused

then we fail here; https://github.com/getsentry/arroyo/blob/ba85564a91f95b0188d626dd3c03bb855f15726c/arroyo/processing/processor.py#L352-L354

I wonder if we can just remove the assertion, the StreamProcessor does not really need it i think

lynnagara:
The rebalance explanation doesn't make sense to me because the line immediately following `consumer.pause() is https://github.com/getsentry/arroyo/blob/5a74ac5b489857de502aad9dbb5ebe06d6717b2e/arroyo/processing/processor.py#L417 which sets `is_paused` to false. So on the next call to poll() none of that code block *should* be executing.

I'm down with removing the assertion though.

lynnagara:
Agree we don't need the assertion. Removed in https://github.com/getsentry/arroyo/pull/305

lynnagara:
Superceded by https://github.com/getsentry/arroyo/pull/305

Issue #307, opened 2023-11-21
Automatically record consumer latency in Arroyo


lynnagara:
Currently at Sentry we primarily monitor backlog size as an indication of consumer health, and we get paged if the backlog is too big. This is problematic because backlog size varies wildly across different consumers, and a number that is too big on one consumer can be perfectly fine on another. The goal is to move to consumer latency everywhere.

One way to solve this in Arroyo could be to add a `message_timestamp` property onto the `Message` class which represents the timestamp of the message from the Kafka broker. Then, when offsets are committed, we can diff the message's timestamp with the current time, which we'll then use as our latency metric.

PR #308, opened 2023-11-22
feat: Enable automatic latency recording for consumers

lynnagara:
The goal of this change is to record a common latency metric for all consumers that use Arroyo.

The new metric is called `arroyo.consumer.latency` -- it represents the time in seconds from when the message was initially produced (broker timestamp) to when the consumer successfully committed the offset for a message.

Note that if a consumer is doing either manual batching (not using Arroyo's built in Reduce/Batch) or manual committing of offsets (not using Arroyo's CommitOffsets strategy), then the metric will not be recorded automatically.

Closes https://github.com/getsentry/arroyo/issues/307

### DIFF
--- a/arroyo/processing/strategies/commit.py
+++ b/arroyo/processing/strategies/commit.py
@@ -1,12 +1,14 @@
+import time
 from typing import Any, Optional
 
 from arroyo.processing.strategies.abstract import ProcessingStrategy
 from arroyo.types import Commit, Message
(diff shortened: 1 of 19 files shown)


untitaker (approved):
we may want to followup at some point to record something more accurate for reduced/unfolded messages (instead of "last value wins", for example aggregate avg/max). nice work!

PR #310, opened 2023-12-01
ref: Add more metrics for slow rebalancing

untitaker:
the unitless metrics are good because they are low overhead to measure, but in the case of rebalancing, the overhead does not matter. emit a metric value everytime partitions are assigned/revoked (and how often a callback decides to shutdown/create strategies as a result), and also emit a counter for how many partitions are being assigned


lynnagara (approved):
thanks!

PR #313, opened 2023-12-08
fix(filter): Handle MessageRejected correctly (INC-584)

untitaker:
When joining a FilterStep, it can happen that the upstream strategy
raises MessageRejected. We need to deal with that somehow, and in this
PR I'm just suppressing that exception.

Also fix a severe bug where MessageRejected was propagated through
submit(), but the MessageRejected was for the wrong message, causing
double-submission of messages.

### DIFF
--- a/arroyo/processing/strategies/filter.py
+++ b/arroyo/processing/strategies/filter.py
@@ -3,7 +3,7 @@ import time
 from typing import Callable, MutableMapping, Optional, Union, cast
 
 from arroyo.commit import CommitPolicy, CommitPolicyState
-from arroyo.processing.strategies.abstract import ProcessingStrategy
+from arroyo.processing.strategies.abstract import MessageRejected, ProcessingStrategy
 from arroyo.types import (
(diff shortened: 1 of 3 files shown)


arroyo/processing/strategies/filter.py
```diff
@@ -86,19 +89,27 @@ def submit(
...
 
-        policy = self.__commit_policy_state
-
-        if policy is not None and policy.should_commit(now, message.committable):
-            self.__flush_uncommitted_offsets(now)
+        if policy is not None and policy.should_commit(now, self.__uncommitted_offsets):
+            # We cannot let MessageRejected propagate here. The caller will
+            # think it is for `message` (which has already been successfully
+            # submitted), and will double-send it.
+            self.__flush_uncommitted_offsets(now, can_backpressure=False)
```
lynnagara, line 96: I think this can be removed since you flush on line 78-79 now

lynnagara (approved):

PR #315, opened 2023-12-10
fix(multiprocessing): Reset pool if tasks are not completed

lynnagara:
If multiprocessing tasks are not completed within the timeout specified, we need to reset the pool to avoid state being carried over between assignments.


untitaker (commented):
i think this change also needs a test. i am also curious if we can reproduce locally whether running into the join timeout "poisons" the pool.

untitaker:
also please add a counter metric when the pool is being recreated, i imagine that with this change, we will have to fine-tune join timeout again to improve consumer performance.

untitaker (approved):
approving since we can't figure out a repro case locally right now, and the test I proposed already exists.

PR #317, opened 2023-12-20
fix(processor): Create strategy lazily on first message

untitaker:
When a series of rebalance events occurs, this avoids recreating the
strategy from within callbacks entirely (making on_revoke in particular
faster). Instead the strategy is created whenever the first message is
being processed.

There were ideas to create the strategy eagerly before even subscribing
to a topic. However, currently create_with_partitions requires knowledge
of the currently assigned partitions. We would have to deprecate this
argument if we wanted to make this happen.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -63,10 +63,6 @@ def _rdkafka_callback(metrics: MetricsBuffer) -> Callable[[F], F]:
     return decorator
 
 
-class InvalidStateError(RuntimeError):
-    pass
-
-
 ConsumerTiming = Literal[
     "arroyo.consumer.poll.time",
     "arroyo.consumer.processing.time",
@@ -210,23 +206,6 @@ class StreamProcessor(Generic[TStrategyPayload]):
 
             self.__metrics_buffer.incr_timing("arroyo.consumer.shutdown.time", value)
 
-        def _create_strategy(partitions: Mapping[Partition, int]) -> None:
-            start_create = time.time()
-
-            self.__processing_strategy = (
-                self.__processor_factory.create_with_partitions(
-                    self.__commit, partitions
-                )
-            )
-
-            self.__metrics_buffer.metrics.timing(
-                "arroyo.consumer.run.create_strategy", time.time() - start_create
-            )
-
-            logger.debug(
-                "Initialized processing strategy: %r", self.__processing_strategy
-            )
-
         @_rdkafka_callback(metrics=self.__metrics_buffer)
         def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
             logger.info("New partitions assigned: %r", partitions)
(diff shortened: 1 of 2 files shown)


arroyo/processing/processor.py
```diff
@@ -243,7 +222,6 @@ def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
                         "Partition assignment while processing strategy active"
                     )
                     _close_strategy()
-                _create_strategy(partitions)
```
lynnagara, line 246: Should we assert that `self.__processing_strategy` is None here? What would happen if we get an assign without a prior revocation? I know that doesn't actually happen today but the code right now is designed to work for incremental assignment as well, which might not be the case after this change.

arroyo/processing/processor.py
```diff
@@ -256,24 +234,6 @@ def on_partitions_revoked(partitions: Sequence[Partition]) -> None:
...
-                            if partition not in partitions
-                        }
-                        logger.info(
-                            "Recreating strategy since there are still active partitions: %r",
-                            active_partitions,
-                        )
-                        _create_strategy(active_partitions)
-                except RuntimeError:
-                    pass
-
```
lynnagara, line 276: Just a note that we don't get a subset of partitions revoked today as we are using eager rebalancing. But that may change in the future. This code was written this way so it also works with incremental rebalancing, even though we never actually recreate the strategy on partitions revoked.

arroyo/processing/processor.py
untitaker: not sure i follow, we do close it on the line before exactly when the strategy is not None

untitaker: fwiw this part of rebalancing was actually broken with incremental assign, as we would create the strategy with the subset of partitions from this callback instead of the full assignment. with this pr we're now actually only calling `_create_strategy` with `consumer.tell()` which should always contain the full assignment

arroyo/processing/processor.py
untitaker: yes i think we're now still handling that case correctly. this code effectively moved from the callback into `_run_once`. both times we recreate the strategy with the return value from `consumer.tell()`. we're still closing and flushing the strategy if any partition had been revoked.

just the condition under which we create the strategy changed. before, we'd recreate the strategy if we still had partitions assigned, now we're creating the strategy if we have none and `consumer.poll()` returned a message.

arroyo/processing/processor.py
```diff
@@ -398,58 +375,59 @@ def _run_once(self) -> None:
             self.__metrics_buffer.incr_timing(
                 "arroyo.consumer.processing.time", time.time() - start_poll
             )
-            if self.__message is not None:
```
fpacifici: I think you need a pretty substantial rebase regarding everything below this line.
I'll review after that.

fpacifici (commented):
What about a slightly more radical idea:
In Arroyo we always provided this strategy factory to make managing the rebalancing indistinguishable from the start/stop of the consumer, thus making the mental model of the user simpler.

Now this, at high scale with a lot of rebalancing, is clearly a liability as well.

What if we made the strategy a mutable concept with these methods:
- submit, poll, join behave as today
- assign(partitions). tells the strategy of the partitions it can use
- revoke(partitions). tells the strategy of the partitions it lost 
- shutdown. Tells the strategy to tear everything down. 

So we just stop rebuilding resources at every rebalance.
Then we can provide an adapter to make the old strategies work in the new model.

Overall I do not have anything specific against this change, please rebase then I will review again.

PR #319, opened 2023-12-28
fix(multiprocessing): Increase default block sizes

nikhars:
There are some Sentry topics for which we allow a single message to be as large as 25 MB. In cases like these, using the autoresizing of block sizes could fail if the default block sizes are smaller than the maximum size of a single message. This change modifies the default block sizes to be atleast as high as the maximum size of a single message.

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -37,8 +37,8 @@ __all__ = ["RunTaskWithMultiprocessing"]
 TResult = TypeVar("TResult")
 TBatchValue = TypeVar("TBatchValue")
 
-DEFAULT_INPUT_BLOCK_SIZE = 16 * 1024 * 1024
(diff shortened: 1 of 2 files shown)


lynnagara:
Were you able to confirm all sentry and snuba consumers have adequate memory requested for this?

untitaker:
for the tests, i had to refactor quite a bit last time to get them to pass: https://github.com/getsentry/arroyo/pull/306

i would recommend fixing the input block size to the old value:


```
@pytest.fixture(autouse=True)
def input_block_size_for_tests(monkeypatch):
    from arroyo.processing.strategies import run_task_with_multiprocessing as x 
    monkeypatch.setattr(x, "DEFAULT_INPUT_BLOCK_SIZE", 16 * 1024 * 1024)
    monkeypatch.setattr(x, "DEFAULT_OUTPUT_BLOCK_SIZE", 16 * 1024 * 1024)

volokluev:

PR #322, opened 2024-01-12
fix(multiprocessing): Implement better error messages for block overflow

untitaker:
See INC-594 -- in a later TT role-playing exercise, it became apparent
that this error message is suboptimal, in that it neither tells you
whether input or output is overflowing, and what parameter to tweak


MeredithAnya (approved):
🎉

PR #328, opened 2024-01-17
fix(rust): Add required metadata for crates.io

untitaker:


untitaker:
we are blocked here on rdkafka actually releasing https://github.com/fede1024/rust-rdkafka/pull/636 on crates.io before we can publish this crate on crates.io. Until then, we can only depend on arroyo as a git dependency as well

PR #331, opened 2024-01-19
test(kafka): Add test for partial cluster failure

untitaker:
For INC-599, in followup SNS-2603, add a test that ensures that the
consumer reliably crashes when a broker dies. I will work a bit more on
the CI setup, but basically you need to run `docker-compose up`, run
that one test using `pytest  tests/backends/test_kafka.py -k died -s`,
and observe that it passes.

This is a problem though because it means we cannot reproduce the
failure we have seen in that incident, where the consumer hangs after
broker failure.


fpacifici (commented):
I don't have anything specific against this PR. 

I think the fact we did not reproduce is more concerning. Do we see any librdkafka logs claiming that we cannot connect during the issue?

untitaker:
this didn't end up going anywhere, closing.

PR #333, opened 2024-01-22
Add flexible Reduce strategy type

cmanallen:
The pull is documented in the code.  Its more contextual to read the reasoning there.  Mostly the buffer protocol and the flush method.  The buffer impl is totally independent of Arroyo/Kafka all the rest.  You can write a buffer with no knowledge of where it will be used.

- Adds the concept of a Buffer protocol.
- Moves all Kafka specific behavior out of the buffer protocol and into the strategy class.
- Allows for flexible buffering without constraint.

### DIFF
--- a/arroyo/processing/strategies/buffer.py
+++ b/arroyo/processing/strategies/buffer.py
@@ -42,7 +42,7 @@ class BufferProtocol(Protocol[TPayload, TResult]):
         ...
 
     @property
-    def ready_to_commit(self) -> bool:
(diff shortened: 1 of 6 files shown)


untitaker (commented):
this mostly looks good, two suggestions:

* remove all references to committing, replace with flushing. we don't know what the next step does
* reimplement Reduce in terms of this, for a lot of additional test coverage

untitaker (approved):

PR #338, opened 2024-02-08
ref(metrics): Add pause/resume counters [INC-626]

untitaker:
As part of the linked incident we think that the ingest consumer
paused/resumed in very quick succession. We currently don't have any
metrics for that.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -81,6 +81,8 @@ ConsumerTiming = Literal[
 ConsumerCounter = Literal[
     "arroyo.consumer.run.count",
     "arroyo.consumer.invalid_message.count",
+    "arroyo.consumer.pause",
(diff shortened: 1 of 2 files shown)


tests/processing/test_processor.py
```diff
@@ -150,7 +150,9 @@ def test_stream_processor_lifecycle() -> None:
         (Timing, "arroyo.consumer.shutdown.time"),
         (Timing, "arroyo.consumer.callback.time"),
         (Timing, "arroyo.consumer.poll.time"),
+        (Increment, "arroyo.consumer.pause"),
```
untitaker, line 153: random thought: i think this test is kind of bs and annoying, but at teh same time it's somewhat interesting to see that we apply backpressure when a consumer is shutting down

lynnagara (approved):

PR #339, opened 2024-02-09
feat(metrics_buffer): Add counters support

phacops:


anonrig:
Can you write a description?

untitaker:
we don't need it, in fact we should IMO strip MetricsBuffer down as much as possible. timers are still there because we cannot aggregate them transparently

PR #340, opened 2024-02-23
feat: Increase backpressure threshold to 5 seconds

lynnagara:
During inc-626 we saw excessive backpressure which caused the consumer to be paused and resumed a lot. This led to excessive network traffic as the local queue was purged and messages were re-downloaded. Increase the threshold for pausing to reduce the impact of this.

### DIFF
--- a/tests/processing/test_processor.py
+++ b/tests/processing/test_processor.py
@@ -86,8 +86,10 @@ def test_stream_processor_lifecycle() -> None:
     with assert_changes(lambda: int(consumer.pause.call_count), 0, 1):
         processor._run_once()
         assert strategy.submit.call_args_list[-1] == mock.call(message)
-        time.sleep(5)
-        processor._run_once()  # Should pause now
+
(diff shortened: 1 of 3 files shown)


tests/processing/test_processor.py
```diff
@@ -86,7 +86,7 @@ def test_stream_processor_lifecycle() -> None:
     with assert_changes(lambda: int(consumer.pause.call_count), 0, 1):
         processor._run_once()
         assert strategy.submit.call_args_list[-1] == mock.call(message)
-        time.sleep(1)
+        time.sleep(5)
```
lynnagara, line 89: 🤮

untitaker: please mock `time.time` instead

untitaker (approved):

tests/processing/test_processor.py
lynnagara: fixed

PR #346, opened 2024-03-27
ref(reduce): Allow to reduce to non-cloneable values

untitaker:
The ergonomics are slightly worse, but I am in a position right now
where the value absolutely cannot be cloneable, and working around those
trait constraints would be even less ergonomical.


Swatinem (approved):
this indeed looks ugly :-(
Curious to see why you would need to resort to not having this be `Clone`.

untitaker:
@Swatinem We'll see, maybe I am going to revert this PR, but I am pretty sure it's for the better. I can also imagine there to be some ergonomic shortcuts for the `Clone` usecase

PR #350, opened 2024-04-05
feat(header): Implement find method on headers

nikhars:
Add the ability in the Headers struct of types to be able to find a key and if found return the corresponding value.

### DIFF
--- a/rust-arroyo/src/backends/kafka/types.rs
+++ b/rust-arroyo/src/backends/kafka/types.rs
@@ -24,12 +24,8 @@ impl Headers {
         Self { headers }
     }
 
-    pub fn find(self, key: &str) -> Option<Vec<u8>> {
-        if let Some(header) = self.headers.iter().find(|header| header.key == key) {
-            header.value.map(|v| v.to_vec())
-        } else {
-            None
-        }
+    pub fn get(self, key: &str) -> Option<&[u8]> {
+        self.headers.iter().find(|header| header.key == key).and_then(|header| header.value.as_ref())
     }
 }
 
--- a/rust-arroyo/src/backends/kafka/types.rs
+++ b/rust-arroyo/src/backends/kafka/types.rs
@@ -1,7 +1,7 @@
 use crate::types::TopicOrPartition;
+use rdkafka::message::Headers as _;
 use rdkafka::message::{BorrowedHeaders, Header, OwnedHeaders};
 use rdkafka::producer::BaseRecord;
-use rdkafka::message::Headers as _;
 
 use std::sync::Arc;
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


rust-arroyo/src/backends/kafka/types.rs
```diff
@@ -22,6 +23,14 @@ impl Headers {
...
         Self { headers }
     }
+
+    pub fn find(self, key: &str) -> Option<Vec<u8>> {
+        if let Some(header) = self.headers.iter().find(|header| header.key == key) {
+            header.value.map(|v| v.to_vec())
+        } else {
+            None
+        }
+    }
```
Swatinem, line 33:
```suggestion
    pub fn get(self, key: &str) -> Option<&[u8]> {
        self.headers.iter().find(|header| header.key == key).and_then(|header| header.value.as_ref())
    }
```

I would suggest to name this `get`, as that is usually used for `HashMap`-like functions.
Also returning `&[u8]` there, as there shouldn’t be a reason to clone inside the function. The caller can do that if necessary.

rust-arroyo/src/backends/kafka/types.rs
```diff
@@ -1,6 +1,7 @@
 use crate::types::TopicOrPartition;
 use rdkafka::message::{BorrowedHeaders, Header, OwnedHeaders};
 use rdkafka::producer::BaseRecord;
+use rdkafka::message::Headers as HeaderTrat;
```
Swatinem, line 4:
```suggestion
use rdkafka::message::Headers as _;
```

If you do not want to *name* traits directly, you can import them as `_` to avoid naming conflicts.

rust-arroyo/src/backends/kafka/types.rs
```diff
@@ -143,4 +151,16 @@ mod tests {
...
+
+    #[test]
+    fn test_headers() {
+        let mut headers = self::Headers::new();
+        headers = headers.insert("key1", Some(b"value1".to_vec()));
+        headers = headers.insert("key2", Some(b"value2".to_vec()));
+        headers = headers.insert("key3", Some(b"value3".to_vec()));
+
+        assert_eq!(headers.get("key1"), Some(b"value1".to_vec()).as_deref());
+        assert_eq!(headers.get("key2"), Some(b"value2".to_vec()).as_deref());
```
Swatinem, line 163:
```suggestion
        assert_eq!(headers.get("key1"), Some(b"value1"));
        assert_eq!(headers.get("key2"), Some(b"value2"));
```

Swatinem (approved):

rust-arroyo/src/backends/kafka/types.rs
nikhars: The suggested change gives the following error
error[E0308]: mismatched types
   --> rust-arroyo/src/backends/kafka/types.rs:162:41
    |
162 |         assert_eq!(headers.get("key1"), Some(b"value1"));
    |                                         ^^^^^^^^^^^^^^^ expected `Option<&[u8]>`, found `Option<&[u8; 6]>`
    |
    = note: expected enum `Option<&[u8]>`
               found enum `Option<&[u8; 6]>`

error[E0308]: mismatched types
   --> rust-arroyo/src/backends/kafka/types.rs:163:41
    |
163 |         assert_eq!(headers.get("key2"), Some(b"value2"));
    |                                         ^^^^^^^^^^^^^^^ expected `Option<&[u8]>`, found `Option<&[u8; 6]>`
    |
    = note: expected enum `Option<&[u8]>`
               found enum `Option<&[u8; 6]>`
               
Didn't know rust would be so restrictive. Made a change to map the value and return its ref.

PR #351, opened 2024-04-09
rust: add more rust logging

dbanda:
Adds more logging to rust to more closely match python. Specifically, we add logging to show when the processor is terminating or closing/rejoining strategies.

We also introduce logging to show the member id as supplied by the broker. This is useful for debugging issue related to rebalancing in kafka since the broker logs only show member id.

### DIFF
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -628,6 +628,11 @@ class KafkaConsumer(Consumer[KafkaPayload]):
     def closed(self) -> bool:
         return self.__state is KafkaConsumerState.CLOSED
 
+    @property
+    def member_id(self) -> str:
+        member_id: str = self.__consumer.memberid()
+        return member_id
+
 
 class KafkaProducer(Producer[KafkaPayload]):
     def __init__(self, configuration: Mapping[str, Any]) -> None:
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -233,6 +233,7 @@ class StreamProcessor(Generic[TStrategyPayload]):
         @_rdkafka_callback(metrics=self.__metrics_buffer)
         def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
             logger.info("New partitions assigned: %r", partitions)
+            logger.info("Member id: %r", self.__consumer.member_id)
             self.__metrics_buffer.metrics.increment(
                 "arroyo.consumer.partitions_assigned.count", len(partitions)
             )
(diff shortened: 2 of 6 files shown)


arroyo/backends/kafka/consumer.py
```diff
@@ -628,6 +628,11 @@ def close(self, timeout: Optional[float] = None) -> None:
     def closed(self) -> bool:
         return self.__state is KafkaConsumerState.CLOSED
 
+    @property
+    def member_id(self) -> str:
+        member_id: str = self.__consumer.memberid()
+        return member_id
```
anonrig, line 634: Does this not work?

```suggestion
        return self.__consumer.memberid()
```

arroyo/processing/processor.py
```diff
@@ -422,7 +423,8 @@ def _run_once(self) -> None:
                         self.__backpressure_timestamp = time.time()
 
                     elif not self.__is_paused and (
-                        time.time() - self.__backpressure_timestamp > BACKPRESSURE_THRESHOLD
+                        time.time() - self.__backpressure_timestamp
+                        > BACKPRESSURE_THRESHOLD
```
anonrig, line 427: This change seems irrelevant?

rust-arroyo/src/backends/kafka/mod.rs
```diff
@@ -505,6 +515,7 @@ mod tests {
     impl AssignmentCallbacks for EmptyCallbacks {
         fn on_assign(&self, partitions: HashMap<Partition, u64>) {
             println!("assignment event: {:?}", partitions);
+
```
anonrig, line 518:
```suggestion
```

arroyo/backends/kafka/consumer.py
dbanda: > Does this not work?

the type checker doesn't like it because `self.__consumer.memberid()` returns `Any`

arroyo/backends/kafka/consumer.py
```diff
@@ -628,6 +628,11 @@ def close(self, timeout: Optional[float] = None) -> None:
     def closed(self) -> bool:
         return self.__state is KafkaConsumerState.CLOSED
 
+    @property
+    def member_id(self) -> str:
```
anonrig, line 632: Maybe we should also add `@cached_property`, depending on the cost?

```suggestion
    @cached_property
    def member_id(self) -> str:
```

anonrig (approved):

arroyo/backends/kafka/consumer.py
dbanda: I think it cheap. Also, I believe the member id changes depending on the rebalancing strategy used. So caching might be inacurate.

Swatinem (approved):
not too happy to see `unsafe` and `unwrap`, but I guess rdkafka does not provide a safe function for this?

lynnagara (approved):

PR #352, opened 2024-04-10
metric(multiprocessing): Add metric to track time it takes to prepare and submit batch 

ayirr7:


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -608,6 +611,11 @@ def __submit_batch(self, input_block_too_small: bool) -> None:
         )
         self.__batch_builder = None
 
+        if self.__last_batch_time == 0.0:
+            self.__last_batch_time = time.time()
+        else:
+            self.__metrics.timing("arroyo.strategies.run_task_with_multiprocessing.batch.build_and_submit_time", time.time() - self.__last_batch_time)
```
nikhars, line 617: I think you also need to reset self.__last_batch_time here.

ayirr7: Oops forgot that line. I think we need to reset it to the current time (so that we have the time that the previous batch was submitted)

arroyo/utils/metric_defs.py
```diff
@@ -98,4 +98,6 @@
     "arroyo.processing.strategies.healthcheck.touch",
     # Number of messages dropped in the FilterStep strategy
     "arroyo.strategies.filter.dropped_messages",
+    # The amount of time it takes for the multiprocessing batch to be built and submitted to subprocesses
+    "arroyo.strategies.run_task_with_multiprocessing.batch.build_and_submit_time"
```
lynnagara, line 102:
```suggestion
    "arroyo.strategies.run_task_with_multiprocessing.batch_build_and_submit_time"
```

PR #356, opened 2024-04-13
feat: A better API for abandoning messages on shutdown

lynnagara:
There were a number of incidents recently (e.g. inc-706) where consumers were not shutdown cleanly as they were unable to finish processing all pending messages before they get kicked out by the broker or kubernetes due to timeouts.

In order to facilitate quicker shutdown, this change aims to make it easier for consumers to drop work and shutdown faster instead of attempting to finish processing every message that has been queued.

A number of strategies that support asynchronous processing (run task in threads, run task with multiprocessing, reduce, unfold, batch, unbatch) now support an optional `abandon_messages_on_shutdown` argument in the constructor. If this is set to true, any work in this processing strategy will be immediately dropped on shutdown. The default is false for backward compatibility.

### DIFF
--- a/arroyo/processing/strategies/run_task_in_threads.py
+++ b/arroyo/processing/strategies/run_task_in_threads.py
@@ -40,6 +40,7 @@ class RunTaskInThreads(
         concurrency: int,
         max_pending_futures: int,
         next_step: ProcessingStrategy[Union[FilteredPayload, TResult]],
+        abandon_messsages_on_shutdown: bool = False,
     ) -> None:
         self.__executor = ThreadPoolExecutor(max_workers=concurrency)
         self.__function = processing_function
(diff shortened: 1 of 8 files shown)


arroyo/processing/strategies/run_task_in_threads.py
```diff
@@ -53,6 +54,7 @@ def __init__(
         self.__next_step = next_step
         self.__closed = False
         self.__invalid_messages = InvalidMessageState()
+        self.__abandon_messages_on_shutdown = abandon_messsages_on_shutdown
```
fpacifici, line 57: Technically join may not only be executed at shutdown. 
I would rename it as abandon messages on join.

fpacifici (changes requested):
Originally the timeout parameter on the join method was supposed to guarantee that we would always complete the join method before the deadline had passed and the consumer was kicked out. 

Is there a reason we cannot use that as we should ? 
Since this will change the public interface of those strategies I think we should not introduce two ways to achieve the same thing: "ensuring the join method returns before the consumer is kicked out"

untitaker:
FYI something similar to this was implemented in https://github.com/getsentry/snuba/pull/5918

volokluev:

PR #359, opened 2024-04-30
ref: Update accumulator sig to return Result<TResult> instead of TResult

john-z-yang:
### Overview

Change the signature for the accumulator in reduce step to return Result<TResult, SubmitError<T>> to generate backpressure.


volokluev (approved):

lynnagara:
can you provide some context about the use-case for this? afaik it's generally used for batching which isn't a backpressure scenario. i believe the python version of this strategy also does not support backpressure  -- is that supposed to be aligned as well?

john-z-yang:
> can you provide some context about the use-case for this? afaik it's generally used for batching which isn't a backpressure scenario. i believe the python version of this strategy also does not support backpressure -- is that supposed to be aligned as well?

In the rust snuba metrics processor, we're using an unbounded queue where we push data on one end and a http client that writes directly to the network buffer on the other end. We're seeing on occasion very high memory usage for the rust processor and we suspect that the rate in which the strategy is processing the messages are faster than the rate in which the http client can write to it, causing the queue to take up a lot of memory.

We want to bound this buffer so that it emits backpressure, but that step is in a reduce step.

rust-arroyo/src/processing/strategies/reduce.rs
```diff
@@ -14,7 +14,9 @@ use super::InvalidMessage;
 
 struct BatchState<T, TResult> {
     value: Option<TResult>,
-    accumulator: Arc<dyn Fn(TResult, T) -> TResult + Send + Sync>,
+    accumulator: Arc<
+        dyn Fn(TResult, Message<T>) -> Result<TResult, (SubmitError<T>, TResult)> + Send + Sync,
```
john-z-yang, line 18: I did not change the signature to borrow TResult exclusively here because I am not sure how we are using this in snuba. But I really prefer that we exclusively borrow TResult here instead of transfering ownership, this way the failure type would be easier to understand.

untitaker: I agree with you, it is probably possible to take `&mut TResult` instead

untitaker (approved):

untitaker:
Since this PR is green, let's merge it. It's a good change regardless of whether we need it.

PR #361, opened 2024-05-14
ref(rust): Remove strategy.close

untitaker:
I long felt that nobody ever does anything useful in this method in Rust
or in Python. And it makes it a bit harder to think about what "state"
the strategy is in,  since strategy.join() sometimes calls close() and
then can no longer be called multiple times.


untitaker:
Also strategy.close is incorrectly implemented in rust's RunTaskInThreads. it closes the next strategy before it does its own flushing, which leads to this sequence of events on rebalancing:

1. self.close
2. next_step.close
3. self.join
4. next_step.submit x 100
5. next_step.join

according to the API contract, `next_step` could panic at step 4 because close was already called. luckily nobody ever implemented that in Rust

loewenheim (approved):

volokluev (approved):

lynnagara (approved):

PR #371, opened 2024-05-21
fix: Allow sending committable from Unfold

### LINKED ISSUE #369: Unfold removes committable from values in ValueBatch
Hi,
I’m not sure if this is a really a bug, but I stumbled over the Unfold strategy’s message generation. Consider the use case where our streaming process consists of the following steps:  
```
0) consume -> 1) batch -> 2) process in batch-> 3) unbatch -> 4) process message wise -> 5) commit
```

When the `ValuesBatch` is unbatched in step 3, the `Unfold` strategy creates a new `Message` instance with a new `Value` which is then submitted to step 4. Unfortunately, only the last message of the batch gets a committable, although all payloads might have one already. From what I gathered from the `BatchStep`’s flow, I assumed it would just fan out the messages and submit them one after the other to the following step, thus unchanged. 

I believe the current behavior, while useful to reduce the number of commits, does not belong to a generic strategy or at least is a bit hidden. My reasoning for moving this out of the strategy or add a note to both classes, would be that for downstream steps like number 4 it is now impossible to provide Partition and offset when raising an `InvalidMessage` exception. Furthermore, in case of an exception in the batches last message, the commit information is lost all together. 

### Environment
arroyo 2.17.4

### Steps to Reproduce

```python
from datetime import datetime
from unittest.mock import MagicMock

from arroyo.backends.kafka import KafkaPayload
from arroyo.processing.strategies import BatchStep, UnbatchStep
from arroyo.types import BrokerValue, Message, Partition, Topic

topic = Topic("test")
partition = Partition(topic, 0)


def test_unfold():
    commit = MagicMock()

    unbatch = UnbatchStep(next_step=commit)

    batch = BatchStep(max_batch_size=2, max_batch_time=3, next_step=unbatch)

    for i in range(1, 6):
        batch.submit(
            message=Message(
                BrokerValue(
                    KafkaPayload(None, b"1", []),
                    Partition(topic, 0),
                    i,
                    datetime.now(),
                )
            )
        )

    # last messages in batch - will receive committable
    assert commit.method_calls[1].args[0].committable == {
        Partition(topic=topic, index=0): 3
    }
    assert commit.method_calls[3].args[0].committable == {
        Partition(topic=topic, index=0): 5
    }

    # this will raise
    assert commit.method_calls[2].args[0].committable == {
        Partition(topic=topic, index=0): 4
    }

```

mj0nez:
This PR aligns the `Unbatch` and `Unfold` strategies to the behavior of their counterparts `Batch` and `Reduce`. 
`Unbatch` now just submits the values of a submitted `ValuesBatch` one after another to the next step. Therefore, the logic  was reduced to rely fully on the passed generator function for building a `Message`'s `Value`.   

I tried my best fixing the typing issues but there is still one in `arroyo/processing/strategies/batching.py:116` with which I had no luck.

Closes  #369

### DIFF
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -112,8 +112,8 @@ class UnbatchStep(
     ) -> None:
         def generator(
             values: ValuesBatch[TStrategyPayload],
-        ) -> MutableSequence[TStrategyPayload]:
-            return [value for value in values]
+        ) -> ValuesBatch[TStrategyPayload]:
+            return values
 
         self.__unfold_step = Unfold(generator, next_step)
 
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -48,14 +48,14 @@ class BatchStep(ProcessingStrategy[Union[FilteredPayload, TStrategyPayload]]):
             result.append(value)
             return result
 
-        self.__reduce_step: Reduce[TStrategyPayload, ValuesBatch[TStrategyPayload]] = (
-            Reduce(
-                max_batch_size,
-                max_batch_time,
-                accumulator,
-                lambda: [],
-                next_step,
-            )
+        self.__reduce_step: Reduce[
+            TStrategyPayload, ValuesBatch[TStrategyPayload]
+        ] = Reduce(
+            max_batch_size,
+            max_batch_time,
+            accumulator,
+            lambda: [],
+            next_step,
         )
 
     def submit(
--- a/arroyo/processing/strategies/batching.py
+++ b/arroyo/processing/strategies/batching.py
@@ -48,14 +48,14 @@ class BatchStep(ProcessingStrategy[Union[FilteredPayload, TStrategyPayload]]):
             result.append(value)
             return result
 
-        self.__reduce_step: Reduce[
-            TStrategyPayload, ValuesBatch[TStrategyPayload]
-        ] = Reduce(
-            max_batch_size,
-            max_batch_time,
-            accumulator,
-            lambda: [],
-            next_step,
+        self.__reduce_step: Reduce[TStrategyPayload, ValuesBatch[TStrategyPayload]] = (
+            Reduce(
+                max_batch_size,
+                max_batch_time,
+                accumulator,
+                lambda: [],
+                next_step,
+            )
         )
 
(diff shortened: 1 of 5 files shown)


arroyo/processing/strategies/batching.py
```diff
@@ -113,7 +113,7 @@ def __init__(
         def generator(
             values: ValuesBatch[TStrategyPayload],
         ) -> MutableSequence[TStrategyPayload]:
-            return [value.payload for value in values]
```
untitaker, line 116: ah yeah, that's why it wasn't fixed. the types need to change then and it's a breaking change to the API. i'm fine with that

arroyo/processing/strategies/unfold.py
```diff
@@ -48,18 +48,13 @@ def submit(self, message: Message[Union[FilteredPayload, TInput]]) -> None:
...
-            # Last message is special because offsets can be committed along with it
-            if idx == num_messages - 1:
-                next_message = Message(
-                    Value(value, message.committable, message.timestamp)
-                )
-            else:
-                next_message = Message(Value(value, {}, message.timestamp))
+        for value in iterable:
+            next_message = Message(
+                value=cast(Value[TOutput], value),
```
untitaker, line 56: I don't think this cast should be necessary, more likely somewhere else the types weren't updated

arroyo/processing/strategies/batching.py
untitaker: since it's no longer the payload you're returning here, the return value needs to change from `MutableSequence[TStrategyPayload]` to `ValuesBatch`. and `[value for value in values]` is probably equivalent to just returning `values`

untitaker:
@mj0nez I haven't looked at the PR in detail yet since your recent changes, do you need help with the types, or what is the status? I can make some room this week to continue it if so

mj0nez:
@untitaker  Did you get a chance to look at the PR? :)

untitaker:
@mj0nez I looked at it for a few hours but couldn't make sense of the errors either. @onkar said at some point he might take a look. sorry for the huge delay, in the end this isn't a huge priority for us and resources are tight.

untitaker (approved):

untitaker:
discard my earlier comment. this is done now. thanks!

PR #375, opened 2024-07-01
Update RunTask to receive a Message<TPayload>

evanpurkhiser:
RunTask now more closely matches the python interface

### DIFF
--- a/rust-arroyo/examples/transform_and_produce.rs
+++ b/rust-arroyo/examples/transform_and_produce.rs
@@ -21,7 +21,8 @@ use rust_arroyo::types::{Message, Topic, TopicOrPartition};
 
 use std::time::Duration;
 
-fn reverse_string(value: KafkaPayload) -> Result<KafkaPayload, InvalidMessage> {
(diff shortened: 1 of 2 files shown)


rust-arroyo/examples/transform_and_produce.rs
```diff
@@ -33,7 +34,7 @@ fn reverse_string(value: KafkaPayload) -> Result<KafkaPayload, InvalidMessage> {
         value.headers().cloned(),
         Some(result_str.to_bytes().to_vec()),
     );
-    Ok(result)
+    Ok(message.replace(result))
```
lynnagara, line 37: I wonder if we should have the `RunTask` strategy call `message.replace(..)` automatically, and leave the return value here. Wdyt?

evanpurkhiser: Let me see what that might look like. @untitaker had some borrow checker concern I think

lynnagara (approved):

PR #376, opened 2024-07-01
feat: Add Noop processing strategy

evanpurkhiser:
This can be useful when you consume a message and do not wish to commit he offset back to Kafka [1].

[1]: The sentry uptime-checker project has this requirement as it is using Kafka as a mechanism to store configurations. By using log compaction we never actually want to commit a log offset, and want to read the entire log every time.


lynnagara (approved):
thanks!

PR #378, opened 2024-07-07
feat: enhance metrics defs

### LINKED ISSUE #360: Enhance metric definitions
Hi,

Thank you for arroyo! I'm currently adopting it and to integrate the exposed metrics with our Prometheus setup, I had to redefine the metrics in [metric_def.py](https://github.com/getsentry/arroyo/blob/8a90458dd2c7b37a883eb9dc7299937077861654/arroyo/utils/metric_defs.py). It would be nice if we could use some simple structure like the following to give each metric a description and type that may be evaluated by downstream consumers/other backends.

```python
from enum import Enum, auto
from typing import NamedTuple

class MetricType(Enum):
    GAUGE = auto()
    TIMER = auto()
    COUNTER = auto()


class Metric(NamedTuple):
    name: str
    type_: MetricType
    description: str


ARROYO_METRICS = [
    Metric("gauge.a", MetricType.GAUGE, "goes up and down"),
    Metric("counter.b", MetricType.COUNTER, "goes up!"),
]
```
AFAIK it's not possible to generate a `Literal` like the current `MetricName` dynamically, so this is something to consider. But if you accept this request, I would be happy to contribute. :)

mj0nez:
This PR adds a script which parses the implemented metric definitions and saves them as a machine readable file. To allow downstream consumers automatic updates on new Arroyo versions, it was integrated into the package build process and will ship with future versions.

Closes #360

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,8 @@ jobs:
         with:
           python-version: 3.8
       - run: |
+          python scripts/create_metrics_def_json.py
+      - run: |
           pip install wheel
           python setup.py sdist bdist_wheel
       - uses: actions/upload-artifact@v3.1.1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,8 @@ jobs:
           pip install -r requirements-linter.txt
       - name: Run linter
         run: |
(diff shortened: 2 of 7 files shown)


untitaker:
Awesome, thanks!

PR #381, opened 2024-07-30
all-repos: update actions/upload-artifact to v4

joshuarli:
Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
       - run: |
           pip install wheel
           python setup.py sdist bdist_wheel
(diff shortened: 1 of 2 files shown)


mwarkentin:
@joshuarli any context for this change? I'm assuming v4 is backwards compatible?

joshuarli:
> @joshuarli any context for this change? I'm assuming v4 is backwards compatible?

v3 is getting deprecated in a few months. It depends - it won't work if there are multiple upload-artifact actions to the same artifact name, because craft expects a single artifact named github.sha for releases. Ref https://github.com/getsentry/craft/issues/552.

In this case, it's fine.

mwarkentin (approved):

PR #382, opened 2024-08-01
confluent-kafka allows headers with str values

asottile-sentry:
https://github.com/confluentinc/confluent-kafka-python/blob/c32d69dfbc61100ac03117573e46fd28a3ac6965/src/confluent_kafka/src/confluent_kafka.c#L1369-L1414


asottile-sentry:
hmm while this is correct it does seem to maybe make consumers more complicated -- so maybe it's best to standardize on bytes here :/

PR #387, opened 2024-09-25
chore: Remove rdkafka git dependency

untitaker:
We were waiting for https://github.com/fede1024/rust-rdkafka/pull/636 to
be merged, but then didn't follow up on this. Now we can finally
release rust-arroyo on crates.io


untitaker:
I keep forgetting that 0.36.2 only looks like this on the git log, but doesn't actually contain our patch...

PR #388, opened 2024-10-17
ref(rust/run_task): Remove unnecessary boxing, make FnMut, change return type

untitaker:
This will allow you to use mutable state from within RunTask where
previously one couldn't, and remove some boilerplate

Also fix a warning drive-by.

### DIFF
--- a/rust-arroyo/src/processing/strategies/run_task.rs
+++ b/rust-arroyo/src/processing/strategies/run_task.rs
@@ -1,6 +1,6 @@
 use crate::processing::strategies::{
-    merge_commit_request, CommitRequest, InvalidMessage, MessageRejected, ProcessingStrategy,
-    StrategyError, SubmitError,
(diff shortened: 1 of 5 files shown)


rust-arroyo/src/processing/strategies/run_task.rs
```diff
@@ -69,7 +63,7 @@ impl<TPayload, TTransformed: Send + Sync> ProcessingStrategy<TPayload>
             return Err(SubmitError::MessageRejected(MessageRejected { message }));
         }
 
-        let next_message = (self.function)(message).map_err(SubmitError::InvalidMessage)?;
+        let next_message = (self.function)(message)?;
```
lynnagara, line 66: does that mean that function provided by the user also needs to handle backpressure now?

untitaker: no, it means it can emit backpressure (that's the goal, to be able to emit backpressure from runtask, something you can already do in python

evanh (approved):

PR #390, opened 2024-11-06
ref(batching): add compute_batch_size to BatchStep

MeredithAnya:
I wanted to be able to batch by an amount other than just the count of messages. I have this implement in https://github.com/getsentry/snuba/pull/6510/, but it's a lot of copying from arroyo, so it would be nice to have it baked into here. ~I'm not sure if we have a ton of other uses for it, but it also seems like it would be relatively harmless to add since by default the batching works the same as before~

Looks like the rust implementation already has something similar (`compute_batch_size`) added in https://github.com/getsentry/arroyo/commit/a5087b8c715c1602b9bd18e3e6a810a97da0988a

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -143,9 +143,9 @@ class StreamProcessor(Generic[TStrategyPayload]):
         self.__processor_factory = processor_factory
         self.__metrics_buffer = MetricsBuffer()
 
-        self.__processing_strategy: Optional[
-            ProcessingStrategy[TStrategyPayload]
-        ] = None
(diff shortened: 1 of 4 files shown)


untitaker (approved):
looks ok. if i understood correctly it's just the parameter name that's different right? generally i'd prefer to name almost everything the same to reduce mental overhead

MeredithAnya:
> looks ok. if i understood correctly it's just the parameter name that's different right? generally i'd prefer to name almost everything the same between rust and python, to reduce mental overhead

I think yes mostly the same, although it seems like in the rust implementation its required, whereas here it defaults to incrementing by 1, I can change the function name though to be consistent

PR #391, opened 2024-11-21
fix: avoid partition revocation on paused consumers

lynnagara:
it is valid for consumers to be paused for periods of time that exceed the max.poll.interval.ms.

we need to continue polling the consumer in this scenario so that the partitions do not get revoked.

librdkafka will not yield any new messages when polling a paused consumer

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -435,6 +435,10 @@ class StreamProcessor(Generic[TStrategyPayload]):
                         self.__consumer.pause([*self.__consumer.tell().keys()])
                         self.__is_paused = True
 
+                    elif self.__is_paused:
+                        # A paused consumer should still poll periodically to avoid it's partitions
+                        # getting revoked by the broker after reaching thte max.poll.interval.ms
+                        self.__consumer.poll(0.1)
                     else:
                         time.sleep(0.01)
 


arroyo/processing/processor.py
```diff
@@ -435,6 +435,10 @@ def _run_once(self) -> None:
                         self.__consumer.pause([*self.__consumer.tell().keys()])
                         self.__is_paused = True
 
+                    elif self.__is_paused:
+                        # A paused consumer should still poll periodically to avoid it's partitions
+                        # getting revoked by the broker after reaching thte max.poll.interval.ms
```
markstory, line 440:
```suggestion
                        # getting revoked by the broker after reaching the max.poll.interval.ms
```

arroyo/processing/processor.py
```diff
@@ -435,6 +435,11 @@ def _run_once(self) -> None:
                         self.__consumer.pause([*self.__consumer.tell().keys()])
                         self.__is_paused = True
 
+                    elif self.__is_paused:
+                        # A paused consumer should still poll periodically to avoid it's partitions
+                        # getting revoked by the broker after reaching the max.poll.interval.ms
+                        # Polling a paused consumer should never yield a message.
+                        assert self.__consumer.poll(0.1) is None
```
MeredithAnya, line 442: do we not want to sleep in this elif block?

lynnagara: it's already got a sleep for `0.1` in there?

MeredithAnya (approved):

PR #392, opened 2024-11-28
metric: Add partition tag to DLQ buffer capacity metric

ayirr7:
This buffer is tracked per partition. Add partition as a tag to see the breakdown.

### DIFF
--- a/rust-arroyo/src/processing/dlq.rs
+++ b/rust-arroyo/src/processing/dlq.rs
@@ -1,10 +1,9 @@
-use std::borrow::Borrow;
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, HashMap, VecDeque};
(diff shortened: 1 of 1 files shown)


evanh (approved):

rust-arroyo/src/processing/dlq.rs
```diff
@@ -403,6 +404,8 @@ impl<TPayload> BufferedMessages<TPayload> {
             return;
         }
 
+        let cloned = message.clone();
```
untitaker, line 407: let's not clone the entire message. you only need the index (not for example, the entire payload), so re-bind that to a new variable here and use it later.

```
let partition_index = message.partition.index;
```

untitaker (approved):

PR #402, opened 2024-12-10
fix(dlq): Do not clone the message if the DLQ is disabled.

untitaker:
If the DLQ is disabled, we can save cloning of the message entirely,
saving 1 allocation


untitaker:
For additional context, I believe that this is the right interface because in the future I can imagine that the DLQ buffer stores its messages in a completely different way (such as, one large contiguous allocation for all payloads)

lynnagara (approved):

PR #404, opened 2024-12-11
ref(dlq): Move DLQ buffer into DLQ policy

ayirr7:
The goal of this PR is to make all DLQ buffer management happen through DLQ policy.

Specifically:
- Ensure no DLQ buffer gets created if there is no DLQ policy defined
- Adding, removing messages to/from buffer goes through DLQ policy
- Log better errors if there are Invalid Messages but no DLQ policy defined

The next step after this refactor is to clear DLQ buffer during rebalancing. This refactor makes it so that we can access the DLQ buffer within the `on_assign` or `on_revoke` callbacks.

### DIFF
--- a/rust-arroyo/src/processing/dlq.rs
+++ b/rust-arroyo/src/processing/dlq.rs
@@ -701,13 +701,6 @@ mod tests {
         assert!(!state.record_invalid_message(&msg));
     }
 
-    // struct TestFactory {}
-    // impl ProcessingStrategyFactory<String> for TestFactory {
-    //     fn create(&self) -> Box<dyn ProcessingStrategy<String>> {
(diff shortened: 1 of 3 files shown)


untitaker (approved):

lynnagara (approved):

ayirr7:
Gonna add some tests since this refactors how buffered messages is accessed

ayirr7:
Added tests and refactored testutils a bit. Let me know if it makes sense.

untitaker (approved):

ayirr7:
Did some local testing as well to run through various cases (DLQ configured without max buffer limit, with max buffer limit, no DLQ configured at all)

PR #407, opened 2024-12-13
feat: Initial support for cooperative-sticky rebalancing

untitaker:
Fix one bug in StreamProcessor where it assumed the passed assignments
are replacing the old ones.

Our consumer backends mostly work as-is, and are already passing the
right values in callbacks.

### DIFF
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -273,7 +273,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         def assignment_callback(
             consumer: ConfluentConsumer, partitions: Sequence[ConfluentTopicPartition]
         ) -> None:
-            if not partitions and self.__is_incremental:
+            if not partitions:
                 logger.info("skipping empty assignment")
                 return
 
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -161,7 +161,10 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         )
 
         configuration = dict(configuration)
-        self.__assignment_strategy = configuration.get("partition.assignment.strategy")
+        self.__is_incremental = (
+            configuration.get("partition.assignment.strategy") == "cooperative-sticky"
+            or configuration.get("group.protocol") == "consumer"
+        )
         auto_offset_reset = configuration.get("auto.offset.reset", "largest")
 
         # This is a special flag that controls the auto offset behavior for
@@ -270,7 +273,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         def assignment_callback(
             consumer: ConfluentConsumer, partitions: Sequence[ConfluentTopicPartition]
         ) -> None:
-            if not partitions and self.__assignment_strategy == "cooperative-sticky":
+            if not partitions and self.__is_incremental:
                 logger.info("skipping empty assignment")
                 return
 
@@ -460,7 +463,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
             ConfluentTopicPartition(partition.topic.name, partition.index, offset)
             for partition, offset in offsets.items()
         ]
-        if self.__assignment_strategy == "cooperative-sticky":
+        if self.__is_incremental:
             self.__consumer.incremental_assign(partitions)
         else:
             self.__consumer.assign(partitions)
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -161,6 +161,7 @@ class KafkaConsumer(Consumer[KafkaPayload]):
         )
 
         configuration = dict(configuration)
+        self.__assignment_strategy = configuration.get("partition.assignment.strategy")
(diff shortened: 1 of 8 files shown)


arroyo/processing/processor.py
```diff
@@ -238,16 +238,23 @@ def on_partitions_assigned(partitions: Mapping[Partition, int]) -> None:
...
+                self.__dlq_policy.reset_dlq_limits(current_partitions)
+            if current_partitions:
                 if self.__processing_strategy is not None:
-                    logger.exception(
+                    # TODO: for cooperative-sticky rebalancing this can happen
+                    # quite often. we should port the changes to
+                    # ProcessingStrategyFactory that we made in Rust: Remove
+                    # create_with_partitions, replace with create +
+                    # update_partitions
+                    logger.error(
```
lynnagara, line 253: can we downgrade to warning if this is expected to happen on a regular basis?

tests/backends/test_kafka.py
```diff
@@ -245,6 +256,10 @@ def test_consumer_polls_when_paused(self) -> None:
                 assert consumer.paused() == []
 
 
+class TestKafkaStreamsIncrementalRebalancing(TestKafkaStreams):
```
lynnagara, line 259: unused?

tests/processing/test_processor.py
```diff
@@ -306,13 +311,15 @@ def test_stream_processor_create_with_partitions() -> None:
...
+    consumer.tell.return_value = {**offsets_p0}
+
     # Second partition assigned
     offsets_p1 = {Partition(topic, 1): 0}
     assignment_callback(offsets_p1)
 
     create_args, _ = factory.create_with_partitions.call_args
     assert factory.create_with_partitions.call_count == 2
-    assert create_args[1] == offsets_p1
+    assert create_args[1] == {**offsets_p1, **offsets_p0}
```
lynnagara, line 322: was this test change related to your other changes? since there's no cooperative rebalancing here, seems like the assertions should stay the same?

lynnagara (commented):
Looks good!

Were you wanting to merge + publish as-is or do more testing against this branch?

Is it worth including a test that fails against the current cluster but works with newer versions?

arroyo/backends/kafka/consumer.py
```diff
@@ -161,6 +161,7 @@ def __init__(
         )
 
         configuration = dict(configuration)
+        self.__assignment_strategy = configuration.get("partition.assignment.strategy")
```
lynnagara, line 164: sorry i said the wrong thing earlier, this should be `group.protocol`

tests/processing/test_processor.py
untitaker: the mocked return value for `consumer.tell` was wrong, so this had the wrong value. the assignments in this test are actually incremental: first p1 is assigned, then p0, and there's no revocation.

arroyo/backends/kafka/consumer.py
untitaker: after discussing offline i think we can support KIP-848 (`group.protocol`) as well as cooperative-sticky rebalancing. they're the same as far as rdkafka API is concerned. i just can't get it to work right now and might scope it out of this PR if it takes too much time.

arroyo/processing/processor.py
untitaker: done

tests/backends/test_kafka.py
untitaker: no, this actually re-declares all the tests in `TestKafkaStreams`, just re-running them with cooperative-sticky rebalancing

untitaker:
> Is it worth including a test that fails against the current cluster but works with newer versions? Attempting to commit on an existing partition during a rebalance might do it.

can you elaborate on this?

arroyo/backends/kafka/consumer.py
```diff
@@ -269,6 +270,10 @@ def subscribe(
         def assignment_callback(
             consumer: ConfluentConsumer, partitions: Sequence[ConfluentTopicPartition]
         ) -> None:
+            if not partitions and self.__assignment_strategy == "cooperative-sticky":
+                logger.info("skipping empty assignment")
+                return
```
fpacifici, line 275: Why do you need a different logic for partition assignment between cooperative and standard in case of empty assignment? 
I assume you can get an empty assignment in the cooperative rebalancing when, after a rebalancing, your assignments do not change. Is that the scenario where you do not want to touch the existing assignments ?

untitaker: after sleeping on it i agree. i only added this because it made cooperative rebalancing more comprehensible, and wasn't sure of the implications on regular rebalancing. i think we can skip empty assignments regardless of the assignment strategy.

tests/test_dlq.py
```diff
@@ -107,7 +107,7 @@ def test_dlq_policy_wrapper() -> None:
     )
     partition = Partition(topic, 0)
     wrapper = DlqPolicyWrapper(dlq_policy)
-    wrapper.reset_offsets({partition: 0})
+    wrapper.reset_dlq_limits({partition: 0})
```
untitaker, line 110: this rename is just to align with rust btw

lynnagara (approved):

PR #412, opened 2024-12-17
fix: Avoid polling the consumer if it says it's paused

untitaker:
There is a bug in StreamProcessor where self.__is_paused, but at the
same time we poll the consumer and get messages.

      File "sentry/runner/commands/run.py", line 455, in basic_consumer
        run_processor_with_signals(processor, consumer_name)
      File "sentry/utils/kafka.py", line 46, in run_processor_with_signals
        processor.run()
      File "arroyo/processing/processor.py", line 323, in run
        self._run_once()
      File "arroyo/processing/processor.py", line 442, in _run_once
        assert self.__consumer.poll(0.1) is None

Breadcrumbs indicate that rebalancing was happening at that time. Could
it be that we were paused, but the consumer got more partitions
assigned? It doesn't make sense since we don't have cooperative-sticky
rebalancing enabled, let's add more logs and prevent the crash.

related: https://github.com/getsentry/arroyo/pull/391

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -446,10 +446,27 @@ class StreamProcessor(Generic[TStrategyPayload]):
                         self.__is_paused = True
 
                     elif self.__is_paused:
-                        # A paused consumer should still poll periodically to avoid it's partitions
-                        # getting revoked by the broker after reaching the max.poll.interval.ms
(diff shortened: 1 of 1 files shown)


MeredithAnya (approved):

untitaker:
related: #413

lynnagara (approved):

PR #417, opened 2024-12-23
fix: increase dlq buffer size

lynnagara:
for our high throughput consumers (which do >1000 messages/sec), this is too small

it's important not to drop messages as we are going to be using dlq more for stale messages too

fixes SENTRY-FOR-SENTRY-A1Z


mwarkentin:
Should this be a config option?

lynnagara:
it's just a safety to prevent unbounded dlq buffer, i don't think we'd want the user to have to configure it and ideally we'll get rid of it longer term. for now just make the error go away.

untitaker (approved):

PR #423, opened 2025-01-17
chore(producer): Add produce time by topic

untitaker:


.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
       - id: fix-encoding-pragma
         args: ["--remove"]
   - repo: https://github.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 7.1.1
```
untitaker, line 20: random change to fix my devenv

arroyo/backends/kafka/consumer.py
```diff
@@ -674,7 +677,14 @@ def __delivery_callback(
...
         error: KafkaError,
         message: ConfluentMessage,
+        start_produce_time: float,
     ) -> None:
+        self.__metrics.timing(
+            "arroyo.producer.produce_time",
+            time.time() - start_produce_time,
+            tags={"topic": message.topic()},
+        )
+
```
fpacifici, line 687: WE should debounce this. Producing one metric per message will be a problem in talking to veneur it would impact the throughput of the consumer too much.

fpacifici (approved):

arroyo/backends/kafka/consumer.py
```diff
@@ -684,6 +684,7 @@ def __delivery_callback(
         error: KafkaError,
         message: ConfluentMessage,
     ) -> None:
+
```
untitaker, line 687:
```suggestion
```

arroyo/backends/kafka/configuration.py
```diff
@@ -49,6 +49,52 @@ def stats_callback(stats_json: str) -> None:
...
 
+def producer_stats_callback(stats_json: str) -> None:
+    stats = json.loads(stats_json)
+    metrics = get_metrics()
+
+    # Extract broker-level int_latency metrics
+    brokers = stats.get("brokers", {})
+    for broker_id, broker_stats in brokers.items():
+        int_latency = broker_stats.get("int_latency", {})
+        if int_latency:
```
evanh, line 60: In what case would this be `None`?

untitaker: not sure when but i think it's at least theoretically possible: https://github.com/getsentry/relay/blob/04f531e6a0d0b3b65283520808f945b030a9457d/relay-kafka/src/producer/utils.rs#L110

evanh: OK fair enough.

arroyo/backends/kafka/configuration.py
```diff
@@ -49,6 +49,52 @@ def stats_callback(stats_json: str) -> None:
     )
 
 
+def producer_stats_callback(stats_json: str) -> None:
```
fpacifici, line 52: Could we have a unit test for this? Just to ensure that if we make changes we do not make this crash the consumer by adding regressions.

fpacifici (approved):
Please consider adding the unit test and the rtt metric before merging.

arroyo/backends/kafka/configuration.py
```diff
@@ -49,6 +49,52 @@ def stats_callback(stats_json: str) -> None:
...
+    stats = json.loads(stats_json)
+    metrics = get_metrics()
+
+    # Extract broker-level int_latency metrics
+    brokers = stats.get("brokers", {})
+    for broker_id, broker_stats in brokers.items():
+        int_latency = broker_stats.get("int_latency", {})
```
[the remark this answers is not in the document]
untitaker: yeah i'm pretty unsure about it to be honest. "broker roundtrip" doesn't mean to me that a request was fully processed but maybe just network roundtrip time of submitting it. i added RTT + avg in any case

rust-arroyo/src/backends/kafka/producer.rs
```diff
@@ -2,18 +2,65 @@ use crate::backends::kafka::config::KafkaConfig;
...
+use rdkafka::producer::{
+    DeliveryResult, ProducerContext as RdkafkaProducerContext, ThreadedProducer,
+};
+use rdkafka::Statistics;
+use std::time::Duration;
+
+pub struct ProducerContext;
+
+impl ClientContext for ProducerContext {
+    fn stats(&self, stats: Statistics) {
```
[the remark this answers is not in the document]
untitaker: added some unit tests. mocking metrics is "too hard" in rust.

PR #427, opened 2025-03-04
fix(unfold): Do not forward empty committable

untitaker:
In https://github.com/getsentry/arroyo/pull/371 we changed the API for
Unfold to allow for custom committable. However, we also made it
mandatory to figure out a committable at all in order for committing to
work.

The callbacks for unfold do not get access to the original committable,
so it's actually not possible to satisfy the 99% usecase with this.

cc @mj0nez

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: fix-encoding-pragma
         args: ["--remove"]
   - repo: https://github.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 5.0.4
     hooks:
       - id: flake8
         language_version: python3.12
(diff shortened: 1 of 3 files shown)


fpacifici (commented):
I think I am missing some context here. 
Is the lack of a committable in the output of the `Unfold`  the problem you are addressing? If yes why is it a problem and how do people commit after Unfold today ? 

Also this is reverting a behavior introduced several months ago. What is the expected impact ?

untitaker:
>  Is the lack of a committable in the output of the Unfold the problem you are addressing? If yes why is it a problem and how do people commit after Unfold today ?

yes, Unfold by default currently produces no committable, and it is the user's job to provide it. but this doesn't actually work because the user, when using Unfold's API, does not have access to the message's partition or offset, only the message payload.

this was plainly broken and I think the only reason is because nobody is actually using Unfold. metrics indexer for example has its own homebrew version of Unfold vendored.

compromise for me is to reenable the old behavior if the user returns `committable={}` for the unfolded data, which I think is never valid anyway.

PR #435, opened 2025-04-01
ref: Remove cmake feature from rdkafka

untitaker:


untitaker:
not necessary, tried to debug a build failure

PR #438, opened 2025-04-03
ref(produce): Add faster way to produce

untitaker:
When producing a large amount of messages, it can be that the profile
consists only of `Future()` calls, and most of the time is being spent
on running that ctor.

When you look inside of Future.__init__, all it does is create
threading.Condition so that `result()` can be blocking.

However, in the Produce step, we do not actually need synchronized
access to the future, nor do we need to have a blocking result(). We
only need done() and non-blocking result().

Introduce ab abstract interface for futures that is fulfilled by stdlib
Future() and a new class that does not offer performant result().

The new future type is not enabled by default as we have uses of the
`Producer` in sentry that do use the full `Future` interface (see `SingletonProducer`)

### DIFF
--- a/arroyo/backends/abstract.py
+++ b/arroyo/backends/abstract.py
@@ -220,7 +220,7 @@ class SimpleProducerFuture(Generic[T]):
         self.result_exception: Exception | None = None
 
     def done(self) -> bool:
-        return self.result_value is not None and self.result_exception is not None
+        return self.result_value is not None or self.result_exception is not None
 
     def result(self, timeout: float | None = None) -> T:
         if self.result_exception is not None:
--- a/arroyo/backends/abstract.py
+++ b/arroyo/backends/abstract.py
@@ -223,9 +223,6 @@ class SimpleProducerFuture(Generic[T]):
         return self.result_value is not None or self.result_exception is not None
 
     def result(self, timeout: float | None = None) -> T:
-        if self.result_exception is not None:
-            raise self.result_exception
-
         if timeout is not None:
             deadline = time.time() + timeout
         else:
@@ -236,6 +233,8 @@ class SimpleProducerFuture(Generic[T]):
         # the contract. If you really need result with timeout>0, you should
         # use the stdlib future.
         while deadline is None or time.time() < deadline:
+            if self.result_exception is not None:
+                raise self.result_exception
             if self.result_value is not None:
                 return self.result_value
             time.sleep(0.1)
--- a/arroyo/backends/abstract.py
+++ b/arroyo/backends/abstract.py
@@ -1,11 +1,23 @@
 from __future__ import annotations
 
 import logging
+import time
 from abc import ABC, abstractmethod, abstractproperty
 from concurrent.futures import Future
-from typing import Callable, Generic, Mapping, Optional, Sequence, Union
+from typing import (
+    Callable,
+    Generic,
+    Mapping,
+    Optional,
+    Protocol,
+    Sequence,
+    TypeVar,
+    Union,
+)
 
 from arroyo.types import BrokerValue, Partition, Topic, TStrategyPayload
+
+T = TypeVar("T")
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


arroyo/backends/abstract.py
```diff
@@ -175,11 +187,69 @@ def member_id(self) -> str:
...
+    A stub for concurrent.futures.Future that does not construct any Condition
+    variables, therefore is faster to construct.
+    """
+
+    def __init__(self) -> None:
+        self.result_value: T | None = None
+        self.result_exception: Exception | None = None
+
+    def done(self) -> bool:
+        return self.result_value is not None
```
evanh, line 223: Should this also check the exception? E.g. is this future "done" if the exception is set?

arroyo/backends/abstract.py
```diff
@@ -175,11 +187,69 @@ def member_id(self) -> str:
...
+
+        if timeout is not None:
+            deadline = time.time() + timeout
+        else:
+            deadline = None
+
+        while deadline is None or time.time() < deadline:
+            if self.result_value is not None:
+                return self.result_value
+            time.sleep(0.1)
```
evanh, line 237: Do we want to make this interval configurable?

untitaker: added a comment, but basically I only wrote this to get the tests to pass. if this becomes performance sensitive you are supposed to use the stdlib future (and that's why it's the default)

SimpleProduceFuture is only good if you use `result(timeout=0)` or call `result()` when you know the future is `done()` (as we do in produce step)

arroyo/backends/abstract.py
untitaker: good catch

evanh (approved):

arroyo/backends/abstract.py
fpacifici: I have some concerns with this implementation. We are providing an implementation of the Future we ask not to use in prod.
What about not using a Future implementation at all? 
If there was no max buffer size in the producer we could jsut call the `produce` method as much as we wanted. OBviously we cannot do that. We need to wait on the callbacks. Why not just keeping a counter with the number of pending callbacks that gets incremented each time a new message is produced and decremented each time a callback is received. 
When the counter reaches the same size as the buffer we start rejecting messages. 
It seems simpler, safer and does not have this concern.

fpacifici (commented):
Please consider addressing the concerns in the comments or consider reverting.

arroyo/backends/abstract.py
untitaker: then this is a larger refactor where we would have to support two APIs or break the futures-based one. A counter would be nice if the produce step used the confluent producer directly.

tests/backends/test_kafka.py
```diff
@@ -150,9 +150,10 @@ def test_auto_offset_reset_earliest(self) -> None:
                 assert isinstance(value, BrokerValue)
                 assert value.payload.value == b"0"
 
-    def test_auto_offset_reset_latest(self) -> None:
+    @pytest.mark.parametrize("use_simple_futures", [True, False])
+    def test_auto_offset_reset_latest(self, use_simple_futures: bool) -> None:
         with self.get_topic() as topic:
-            with closing(self.get_producer()) as producer:
+            with closing(self.get_producer(use_simple_futures)) as producer:
```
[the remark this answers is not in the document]
untitaker: I will follow up on this

asottile-sentry:
this appears to be a breaking change released in a patch version -- I am unable to bump this library in `sentry` to unbreak a bunch of flaky tests

should this be reverted and reapplied in a more compatible way and / or in a major version bump?

untitaker:
if possible let's roll forward in https://github.com/getsentry/sentry/pull/88694 -- as said on slack I do not believe the semver violation makes a difference here as you would've been blocked by the upgrade either way. for example if I had released arroyo 3.0 you'd have to release 3.1 or 3.0.1 and we'd be in the same situation.

we had this problem before, at inconvenient times. any suggestions welcome.

PR #439, opened 2025-04-03
inc: Add logs to Arroyo to track behavior during paused and empty message states

ayirr7:
We currently don't have these logs, which will also be helpful for breaking down per-consumer

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -35,6 +35,7 @@ logger = logging.getLogger(__name__)
 
 METRICS_FREQUENCY_SEC = 1.0  # In seconds
 BACKPRESSURE_THRESHOLD = 5.0  # In seconds
+LOGGING_FREQUENCY_SEC = 60.0  # In seconds
 
 F = TypeVar("F", bound=Callable[[Any], Any])
 
@@ -168,6 +169,10 @@ class StreamProcessor(Generic[TStrategyPayload]):
         self.__dlq_policy: Optional[DlqPolicyWrapper[TStrategyPayload]] = (
             DlqPolicyWrapper(dlq_policy) if dlq_policy is not None else None
         )
+
+        self.__last_run_log_ts = time.time()
+        self.__last_pause_ts = time.time()
+        self.__last_empty_msg_ts = time.time()
 
         def _close_strategy() -> None:
             start_close = time.time()
@@ -395,6 +400,10 @@ class StreamProcessor(Generic[TStrategyPayload]):
     def _run_once(self) -> None:
         self.__metrics_buffer.incr_counter("arroyo.consumer.run.count", 1)
 
+        if time.time() - self.__last_run_log_ts >= LOGGING_FREQUENCY_SEC:
+            logger.info("Arroyo consumer _run_once loop started")
+            self.__last_run_log_ts = time.time()
+
         message_carried_over = self.__message is not None
 
         if not message_carried_over:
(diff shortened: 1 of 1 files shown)


arroyo/processing/processor.py
```diff
@@ -408,6 +417,12 @@ def _run_once(self) -> None:
                 self.__metrics_buffer.incr_timing(
                     "arroyo.consumer.poll.time", time.time() - start_poll
                 )
+
+                if self.__message is None and not self.__is_paused:
+                    if time.time() - self.__last_empty_msg_ts >= LOGGING_FREQUENCY_SEC:
+                        logger.info("Consumer is not paused, but did not receive a message from underlying consumer")
+                        self.__last_empty_msg_ts = time.time()
```
fpacifici, line 424: I don't think this is right. 
Let's say we get None, then 30 seconds later we get a valid message, we are not resetting the counter. After 60 seconds we would log the message even if we received a message in the 60 seconds.
The goal was to periodically log if we did not received any message for a long period of time.

arroyo/processing/processor.py
```diff
@@ -477,6 +492,10 @@ def _run_once(self) -> None:
                             # getting revoked by the broker after reaching the max.poll.interval.ms
                             # Polling a paused consumer should never yield a message.
                             assert self.__consumer.poll(0.1) is None
+
+                            if time.time() - self.__last_pause_ts >= LOGGING_FREQUENCY_SEC:
+                                logger.info("Consumer is paused, polling")
+                                self.__last_pause_ts = time.time()
```
fpacifici, line 498: Same here, we want to verify that we have been in a paused state without unpausing for long. 
So if we unpause the clock has to be reset.

fpacifici (changes requested):

arroyo/processing/processor.py
```diff
@@ -35,6 +35,7 @@
 
 METRICS_FREQUENCY_SEC = 1.0  # In seconds
 BACKPRESSURE_THRESHOLD = 5.0  # In seconds
+LOGGING_FREQUENCY_SEC = 60.0  # In seconds
```
fpacifici, line 38: Let's make it longer now that we know what the likely issue is. 
64 consumers would log this every second. 
What about once every 2-3 minutes ?

fpacifici (approved):

arroyo/processing/processor.py
```diff
@@ -395,6 +400,10 @@ def _handle_invalid_message(self, exc: InvalidMessage) -> None:
     def _run_once(self) -> None:
         self.__metrics_buffer.incr_counter("arroyo.consumer.run.count", 1)
 
+        if time.time() - self.__last_run_log_ts >= LOGGING_FREQUENCY_SEC:
```
untitaker, line 403: please only calculate `time.time()` once, it's already very expensive to run it per `_run_once`

we also already have a counter metric for this, what is the purpose of the log?

arroyo/processing/processor.py
```diff
@@ -169,6 +170,10 @@ def __init__(
             DlqPolicyWrapper(dlq_policy) if dlq_policy is not None else None
         )
 
+        self.__last_run_log_ts = time.time()  # This is for throttling the logging of each run loop per-consumer
+        self.__last_pause_ts = None
```
untitaker: you can simplify your logic by setting all of these to `0`, then you don't have to check for `None`

arroyo/processing/processor.py
```diff
@@ -408,6 +417,26 @@ def _run_once(self) -> None:
...
                 )
+
+                if self.__message is None:
+                    if not self.__is_paused:
+                        if self.__last_empty_msg_ts is None:
+                            self.__last_empty_msg_ts = time.time()
+
+                        # Records a log if the consumer has been active but receiving no message from poll() for longer than a threshold duration
+                        elif time.time() - self.__last_empty_msg_ts >= LOGGING_FREQUENCY_SEC:
+                            logger.info(f"Consumer is not paused but did not receive a message from underlying consumer for {LOGGING_FREQUENCY_SEC} seconds")
```
untitaker, line 428: seems severe enough to `warn`

arroyo/processing/processor.py
ayirr7: Per-consumer breakdown

PR #440, opened 2025-04-03
avoid leaking SIGCHLD handler in RunTaskWithMultiprocessing

asottile-sentry:
~technically this should probably be a context manager but it seems to use java-like `.close()` pattern instead and I'm not about to try and fix every caller

### DIFF
--- a/arroyo/processing/strategies/run_task_with_multiprocessing.py
+++ b/arroyo/processing/strategies/run_task_with_multiprocessing.py
@@ -581,7 +581,7 @@ class RunTaskWithMultiprocessing(
                 self.__metrics.increment("sigchld.detected")
                 raise ChildProcessTerminated(signum)
 
-        signal.signal(signal.SIGCHLD, handle_sigchld)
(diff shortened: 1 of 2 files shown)


arroyo/processing/strategies/run_task_with_multiprocessing.py
```diff
@@ -839,14 +839,19 @@ def submit(
             # compression.)
             self.__batch_builder.append(message)
 
-    def close(self) -> None:
+    def _do_close(self) -> None:
         self.__closed = True
 
+        signal.signal(signal.SIGCHLD, self.original_sigchld)
```
fpacifici, line 845: I am not sure I get the reason to reinstall the original sigchld handler inside the close method.

asottile-sentry: as written now if this is used in a process (say, getsentry's testsuite) the SIGCHILD handler persists after

in my case it's causing test pollution by a future test being broken by the sigchld handler

fpacifici (approved):

PR #442, opened 2025-04-07
wip: Stuck detector

untitaker:


fpacifici (commented):
Do you have context about this? Is there a specific scenario where the consumer got stuck?

untitaker:
@fpacifici I opened this PR because I was chatting with Riya about https://github.com/getsentry/arroyo/pull/439, but I think I misunderstood the situation there. The consumer gets stuck in paused state, not stuck in a way that it exceeds max-poll-intervall. We've had a few situations where the main thread got stuck, but none recently

PR #444, opened 2025-04-07
inc: Add a unit test that demonstrates InvalidMessage failure mode

ayirr7:
If the consumer is paused, and is carrying over a message, and then that same message ends up being Invalid, the consumer gets stuck forever. It never leaves the paused state. 

This unit test demonstrates this failure mode. A following PR will introduce a fix and an adjustment to this unit test.

### DIFF
--- a/tests/processing/test_processor.py
+++ b/tests/processing/test_processor.py
@@ -657,15 +657,15 @@ def test_processor_pause_with_invalid_message() -> None:
     with assert_changes(lambda: int(consumer.pause.call_count), 0, 1):
         processor._run_once()
         assert strategy.submit.call_args_list[-1] == mock.call(message)
-        assert processor.__message == message.value
+        assert processor.StreamProcessor__message == message.value
 
         with mock.patch("time.time", return_value=time.time() + 5):
             processor._run_once()  # Should pause now
 
     # Consumer is in paused state
-    assert processor.__is_paused is True
+    assert processor.StreamProcessor__is_paused is True  # type: ignore
     # The same rejected message should be carried over
-    assert processor.__message is not None
+    assert processor.StreamProcessor__message is not None  # type: ignore
 
     # All partitions are paused
     consumer.paused.return_value = set(p for p in offsets)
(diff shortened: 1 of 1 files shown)


tests/processing/test_processor.py
```diff
@@ -622,3 +622,87 @@ def test_healthcheck(tmpdir: py.path.local) -> None:
...
+    # Simulate a continuous backpressure state where messages are being rejected
+    strategy.submit.side_effect = MessageRejected()
+
+    # Simulate Kafka returning nothing since the consumer is paused
+    consumer.poll.return_value = None
+
+    # The next poll returns nothing, but we are still carrying over the rejected message
+    processor._run_once()
+    assert consumer.poll.return_value is None
+    assert processor._StreamProcessor__is_paused is True  # type: ignore
```
ayirr7, line 681: I think these assertions here actually really help the user reason about the internals of the processor and helps for making what this test does clear. Unfortunately it requires a # type: ignore but I figured since it's a test and it exists in order to explain a failure mode, it would be ok

tests/processing/test_processor.py
```diff
@@ -622,3 +622,87 @@ def test_healthcheck(tmpdir: py.path.local) -> None:
...
+
+    # Consumer is in paused state
+    assert processor._StreamProcessor__is_paused is True  # type: ignore
+    # The same rejected message should be carried over
+    assert processor._StreamProcessor__message is not None  # type: ignore
+
+    # All partitions are paused
+    consumer.paused.return_value = set(p for p in offsets)
+    # Simulate a continuous backpressure state where messages are being rejected
+    strategy.submit.side_effect = MessageRejected()
```
fpacifici, line 673: I don't think you need to assert these internals of StreamProcessor. 
These are implementation details that make a test more fragile when they leak out.
You can still test the property of the system (break if the consumer is paused and the next message is invalid) without asserting the internals.
Adding the assertions on internals means the test breaks upon refactoring, while having a test passing after a backwards compatible refactoring is an important asset to validate the change.

fpacifici (approved):
Please avoid the assertions on internals unless strictly needed. I don't think they are needed here.

tests/processing/test_processor.py
untitaker: is this on the right line? i think this is ok to induce paused state of the consumer. I agree that `assert is_paused` is not really necessary

PR #445, opened 2025-04-07
inc-fix: Don't make a paused consumer stuck when dealing with InvalidMessage

ayirr7:
A follow-up to https://github.com/getsentry/arroyo/pull/444. PLEASE TAKE A LOOK AT THAT PR FIRST. 

A prior unit test shows that the consumer gets stuck when it is paused _and_ deals with an InvalidMessage (for example, because the carried-over message from backpressure mode is now stale). 

Handle that by unpausing the consumer and fixing the unit test to show the new behavior


arroyo/processing/processor.py
```diff
@@ -425,9 +425,7 @@ def _run_once(self) -> None:
             if self.__message is not None:
                 try:
                     start_submit = time.time()
-                    message = (
```
ayirr7, line 428: IDK why this was here before, let me know if removing this is a mistake

ayirr7:
https://github.com/getsentry/arroyo/blob/main/arroyo/processing/processor.py#L416-L421

There's this other block where we handle InvalidMessages. However, I don't think we need to change anything there. See this path of the message:
- Consumer is paused
- Polling the strategy yields some invalid messages which we handle by producing to DLQ, whatever, etc.
- Return 
- Upon next `_run_once()` call, we're dealing with the same message which was carried over from before the consumer was paused
- We will hold onto that message until we leave backpressure mode

arroyo/processing/processor.py
fpacifici: This is different now. If `self.__message` is None now it would return `Message(None)`. Before it would have returned `None`. Is this change intentional ?

untitaker: i think this change is correct since the entire block only runs if message is not none (see line 425) -- probably a copypaste error or to pacify mypy

untitaker (approved):

arroyo/processing/processor.py
ayirr7: yeah that block does not even run

PR #446, opened 2025-04-08
Revert "ref(producer): Simplify typing, add more tests"

ayirr7:
Reverts getsentry/arroyo#441

Can't bump Arroyo version right now in Sentry as there are some typing issues, we can come back to it afterwards


fpacifici:
Could you please link to the specific type issue so Markus will know what to fix ?

fpacifici (approved):

PR #448, opened 2025-04-09
Reapply "ref(producer): Simplify typing, add more tests"

untitaker:
Reverts getsentry/arroyo#446


jan-auer (approved):
Unblocking the PR. In the long run I would still suggest a second, separate produce step.

PR #451, opened 2025-04-26
feat: add backpressure time limit

davidtsuk:
In order to avoid being in a backpressure state forever, add an option to set a time limit for the amount of time you can spend in backpressure. If we reach the time limit, we panic, I'm not sure if there's a better way to gracefully shut down the consumer.

### DIFF
--- a/rust-arroyo/src/processing/mod.rs
+++ b/rust-arroyo/src/processing/mod.rs
@@ -64,13 +64,12 @@ impl<TPayload: Send + Sync + 'static> ConsumerState<TPayload> {
     pub fn new(
         processing_factory: Box<dyn ProcessingStrategyFactory<TPayload>>,
         dlq_policy: Option<DlqPolicy<TPayload>>,
-        backpressure_timelimit: Option<Duration>,
     ) -> Self {
         let inner = ConsumerStateInner {
             processing_factory,
             strategy: None,
             backpressure_deadline: None,
-            backpressure_timelimit,
+            backpressure_timelimit: None,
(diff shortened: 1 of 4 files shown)


untitaker (approved):
I would avoid adding a new mandatory parameter on StreamProcessor and instead build an API like this:

```
StreamProcessor::with_kafka(..).with_backpressure_timelimit(...)
```

rust-arroyo/src/processing/mod.rs
```diff
@@ -405,6 +411,15 @@ impl<TPayload: Clone + Send + Sync + 'static> StreamProcessor<TPayload> {
...
                 }
+
+                if let Some(backpressure_timelimit) = consumer_state.backpressure_timelimit {
+                    if deadline.elapsed() > backpressure_timelimit {
+                        panic!(
+                            "Consumer is in backpressure state for more than the time limit of {} seconds, panicking",
+                            backpressure_timelimit.as_secs()
+                        );
+                    }
+                }
```
kylemumma, line 422: in the case we panic we have variable `ret` which is a MessageRejected error `Err(SubmitError::MessageRejected(MessageRejected { message })` this is the rejection that caused this specific backpressure request. This would be good to get logged in order to figure out whats causing the backpressure.

But the catch is its only this 1 instance, not the mutiple continuous requests that likely caused it.

kylemumma: `self.message` might be sufficient idk if the object has anything else important

untitaker (approved):

PR #455, opened 2025-05-01
feat: Improve health check

davidtsuk:
Instead of touching the health file every time we call poll, we only touch it when one of the following two conditions is true:
1. We have a commit request which means we are actively processing messages
2. We are calling poll after not calling submit in the previous iteration - this happens when we don't have messages to process which is a healthy state

### DIFF
--- a/rust-arroyo/src/processing/strategies/healthcheck.rs
+++ b/rust-arroyo/src/processing/strategies/healthcheck.rs
@@ -14,6 +14,7 @@ pub struct HealthCheck<Next> {
     path: PathBuf,
     interval: Duration,
     deadline: SystemTime,
+    iterations_since_last_submit: u64,
 }
 
 impl<Next> HealthCheck<Next> {
@@ -26,6 +27,7 @@ impl<Next> HealthCheck<Next> {
             path: path.into(),
             interval,
             deadline,
+            iterations_since_last_submit: 0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


rust-arroyo/src/processing/strategies/healthcheck.rs
```diff
@@ -50,12 +52,21 @@ where
...
+            self.maybe_touch_file();
+        }
 
-        self.next_step.poll()
+        self.iterations_since_last_submit += 1;
+        poll_result
     }
 
     fn submit(&mut self, message: Message<TPayload>) -> Result<(), SubmitError<TPayload>> {
+        self.iterations_since_last_submit = 0;
```
davidtsuk, line 69: not sure if we should check if submit was successful before reseting this counter

rust-arroyo/src/processing/strategies/healthcheck.rs
```diff
@@ -14,6 +14,7 @@ pub struct HealthCheck<Next> {
     path: PathBuf,
     interval: Duration,
     deadline: SystemTime,
+    iterations_since_last_submit: u64,
```
davidtsuk, line 17: I'm using a u64, but maybe we can get away with a u32

evanh (approved):
Can you clarify in your commit message when we would end up in the situation where those two conditions are _not_ met?

Also, can you add tests?

fpacifici:
As this is something you are testing for the Snuba consumer I would do this inside the snuba consumer code rather than the standard HealthCheck in the arroyo code base.
That would allow you to have runtime options to turn this behavior on and off. 
This approach would be a breaking change for other consumers as well.

PR #462, opened 2025-06-03
feat: add consumer_member_id to metrics by default

### LINKED ISSUE #357: Record consumer / client ID on metrics
We currently have a bunch of metrics in the `arroyo` namespace, however they are only tagged with VM / host information, rather than consumer, so it is hard (or impossible) to breakdown metrics properly per consumer.

One example of this would be to try to determine if partitions are balanced evenly across consumers. We have a metric called `arroyo.consumer.partitions_assigned.count` however we can only break it down by instance-id and similar tags which may include partitions assigned to multiple consumers running on the same node:

image

Ideally we could break this down by consumer and see if they all have the same number of partitions assigned.

mwarkentin:
# Add consumer_member_id to metrics by default

Fixes #357 

## Summary

This PR automatically adds `consumer_member_id` tags to all consumer-related metrics, improving observability and debugging capabilities in distributed consumer environments.

### DIFF
--- a/arroyo/processing/processor.py
+++ b/arroyo/processing/processor.py
@@ -29,7 +29,7 @@ from arroyo.processing.strategies.abstract import (
 )
 from arroyo.types import BrokerValue, Message, Partition, Topic, TStrategyPayload
 from arroyo.utils.logging import handle_internal_error
-from arroyo.utils.metrics import get_consumer_metrics, Tags
+from arroyo.utils.metrics import get_consumer_metrics, Tags, MetricName
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


arroyo/processing/processor.py
```diff
@@ -122,6 +123,18 @@ def __throttled_record(self) -> None:
         if time.time() - self.__last_record_time > METRICS_FREQUENCY_SEC:
             self.flush()
 
+    def record_timing(self, metric: MetricName, value: Union[int, float], additional_tags: Optional[Tags] = None) -> None:
```
untitaker, line 126: those wrapper functions don't seem to serve a purpose and don't add additional tags

PR #463, opened 2025-06-13
add a message take method

volokluev:


untitaker (approved):
neat, i think you can remove a use of RefCell in the clickhouse writer step with this

i'd go further and deprecate try_map, up to you if you want to do it in this pr though

fpacifici:
> Also nit: I think the name take is pretty vague for this function since I would expect a function like take to return just a Message rather than a tuple. I think something like into_parts would be a better name and more consistent with the into_payload name.

`take` is a known method in the `Option` struct in rust. This has the same semantics, I think the name comes from there, so it seems appropriate. https://doc.rust-lang.org/std/option/enum.Option.html#method.take

untitaker:
I mean, another option for the name is `into_parts()` which is also commonly used on methods that return a tuple. but i prefer `take()`.

if you want to make it more like `Option::take` i guess you'd have to mutate the original message somehow, but that's not possible since you want to change the payload type...

PR #465, opened 2025-06-25
chore(deps): depend on sentry-core >=0.32

lcian:
This allows us to upgrade the SDK in crates that depend on both the SDK and `arroyo` itself, and not end up with two separate versions in the lockfile and at runtime.
Additionally, we just need to depend on `sentry_core`, as the only functionality we're using here is the `Hub`.

Ref https://github.com/getsentry/snuba/pull/7250

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,14 @@ coarsetime = "0.1.33"
 once_cell = "1.18.0"
 rand = "0.8.5"
 rdkafka = { version = "0.37.0", features = ["cmake-build", "tracing"] }
-sentry = { version = "0.32.0" }
+sentry = { version = "0.41.0", default-features = false, features = [
+    # default features, except `release-health` is disabled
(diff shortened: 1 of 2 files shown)


lcian:
This is not really needed for Snuba as it should work anyway even if the versions are out of sync.

untitaker:
if i merge this PR and bump arroyo in a repo that doesn't have the SDK upgraded, what happens? will we just get a version split?

lcian:
> if i merge this PR and bump arroyo in a repo that doesn't have the SDK upgraded, what happens? will we just get a version split?

Yeah, the versions should be always kept in sync.

untitaker:
> Yeah, the versions should be always kept in sync.

A version split would be OK. it would be worse if we cannot upgrade arroyo without upgrading the SDK.

i'd suggest to relax the version constraint anyway though

PR #470, opened 2025-07-17
fix(slo): Add a metric to capture produce success/failures for our SLO

evanh:
This adds a metric that counts how many successful produces there were with a status. A future PR
will also add the specific error code to the tags.

### DIFF
--- a/arroyo/processing/strategies/produce.py
+++ b/arroyo/processing/strategies/produce.py
@@ -13,6 +13,7 @@ from arroyo.types import (
     TStrategyPayload,
     Value,
 )
+from arroyo.utils.metrics import get_metrics
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


untitaker (approved):

arroyo/processing/strategies/produce.py
```diff
@@ -71,9 +73,20 @@ def poll(self) -> None:
...
+                try:
+                    result = future.result()
+                    self.__metrics.increment(
+                        "arroyo.producer.produce_status", tags={"status": "success"}
+                    )
+                except Exception as e:
+                    self.__metrics.increment(
+                        "arroyo.producer.produce_status", tags={"status": "error"}
+                    )
+                    raise e
```
fpacifici, line 85: We need to identify the error code and add it as a tag. Not all errors should make us burn our SLO. For example messages that are too large would still show up here. Those should not make us burn the SLO.

PR #471, opened 2025-07-17
fix(metrics): Add an error code to the produce SLO

evanh:
If the exception is a KafkaException, add the error code to the metric produced.


evanh:
Need to figure out rust and validate Python.

PR #472, opened 2025-07-18
fix(slo): Add error codes to producer failure in Rust

evanh:
This adds the error codes to the producer failure metric in Rust. Some new producer errors were also
added that explicitly deal with producing (similar to the consumer errors).

### DIFF
--- a/rust-arroyo/src/backends/kafka/producer.rs
+++ b/rust-arroyo/src/backends/kafka/producer.rs
@@ -102,7 +102,7 @@ impl ArroyoProducer<KafkaPayload> for KafkaProducer {
 
         self.producer
             .send(base_record)
-            .map_err(|_| ProducerError::ProducerErrorred)?;
+            .map_err(|(kafka_error, _record)| ProducerError::from(kafka_error))?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


rust-arroyo/src/backends/kafka/producer.rs
```diff
@@ -243,4 +243,20 @@ mod tests {
             .produce(&destination, payload)
             .expect("Message produced")
     }
+
+    #[test]
+    #[ignore = "This should fail with a KafkaError but doesn't for some reason"]
```
evanh, line 248: I could not figure out how to get the produce to fail. Even with a non-existent topic, referencing a partition that doesn't exist, on a broker that doesn't exist, this test still passes.

untitaker: Ah yes, @fpacifici @bmckerry found an issue with these producers: they do not await until the produce is actually done. see how the arroyo produce waits with callback+future, that's missing. maybe you can fix it in the same PR? otherwise it goes to the backlog

evanh: Ah OK I'll see if I can fix that.

PR #473, opened 2025-07-18
fix(produce): Ensure produce waits for buffer flush

evanh:
The rdkafka library has two times when it will return error messages: if the message fails to be added to the produce buffer, and if the message fails to be flushed out of the buffer. Currently only the first case was being handled.

When a message is sent, attach a channel using the `DeliveryOpaque` concept. When the message is flushed from the buffer, the context callback is called. Use that channel to forward on the success or failure to the `send` function.

The `send` function will now wait on the receive channel for the message to be sent. If it takes longer than 5 seconds, a new error `ProduceWaitTimeout` will be sent.

Since this `send` function is already called inside a synchronous thread, this won't block other messages from producing.

Depends on some concepts added in https://github.com/getsentry/arroyo/pull/472

### DIFF
--- a/rust-arroyo/src/backends/kafka/producer.rs
+++ b/rust-arroyo/src/backends/kafka/producer.rs
@@ -6,10 +6,13 @@ use crate::timer;
 use crate::types::TopicOrPartition;
 use rdkafka::client::ClientContext;
 use rdkafka::config::ClientConfig;
+use rdkafka::error::KafkaError;
 use rdkafka::producer::{
     DeliveryResult, ProducerContext as RdkafkaProducerContext, ThreadedProducer,
 };
 use rdkafka::Statistics;
+use std::sync::mpsc::{channel, Sender};
+use std::sync::Arc;
 use std::time::Duration;
 
 pub struct ProducerContext;
@@ -65,13 +68,21 @@ impl ClientContext for ProducerContext {
 }
 
 impl RdkafkaProducerContext for ProducerContext {
-    type DeliveryOpaque = ();
+    type DeliveryOpaque = Arc<Sender<Result<(), KafkaError>>>;
 
     fn delivery(
         &self,
         _delivery_result: &DeliveryResult<'_>,
         _delivery_opaque: Self::DeliveryOpaque,
     ) {
+        let result = match _delivery_result {
+            Ok(_) => Ok(()),
+            Err((err, _)) => Err(err.clone()),
+        };
+
+        _delivery_opaque
+            .send(result)
+            .expect("Failed to send delivery result");
     }
 }
 
@@ -98,13 +109,21 @@ impl ArroyoProducer<KafkaPayload> for KafkaProducer {
         destination: &TopicOrPartition,
         payload: KafkaPayload,
     ) -> Result<(), ProducerError> {
-        let base_record = payload.to_base_record(destination);
+        let (tx, rx) = channel::<Result<(), KafkaError>>();
+        let base_record = payload.to_base_record(destination, Arc::new(tx));
 
+        // If the producer fails to enqueue the message, this will return an error
         self.producer
             .send(base_record)
             .map_err(|(kafka_error, _record)| ProducerError::from(kafka_error))?;
 
-        Ok(())
+        // If the producer fails to flush the message out of the buffer, the delivery callback will send an error
+        // on the channel.
+        match rx.recv_timeout(Duration::from_secs(5)) {
+            Ok(Ok(_)) => Ok(()),
+            Ok(Err(e)) => Err(ProducerError::from(e)),
+            Err(_) => Err(ProducerError::ProduceWaitTimeout),
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


rust-arroyo/src/backends/kafka/producer.rs
```diff
@@ -65,13 +68,21 @@ impl ClientContext for ProducerContext {
 }
 
 impl RdkafkaProducerContext for ProducerContext {
-    type DeliveryOpaque = ();
+    type DeliveryOpaque = Arc<Sender<Result<(), KafkaError>>>;
```
untitaker, line 71: why does this sender need to be arced?

rust-arroyo/src/backends/kafka/producer.rs
```diff
@@ -98,13 +109,21 @@ impl ArroyoProducer<KafkaPayload> for KafkaProducer {
...
             .map_err(|(kafka_error, _record)| ProducerError::from(kafka_error))?;
 
-        Ok(())
+        // If the producer fails to flush the message out of the buffer, the delivery callback will send an error
+        // on the channel.
+        match rx.recv_timeout(Duration::from_secs(5)) {
+            Ok(Ok(_)) => Ok(()),
+            Ok(Err(e)) => Err(ProducerError::from(e)),
+            Err(_) => Err(ProducerError::ProduceWaitTimeout),
+        }
```
untitaker, line 126:
```suggestion
        rx.recv_timeout(Duration::from_secs(5))
             .map_err(|_| ProducerError::ProduceWaitTimeout)?
             .map_err(ProducerError::from)
```

(untested)

untitaker: but also not sure, do we want this 5 second timeout?

rust-arroyo/src/backends/kafka/producer.rs
evanh: In order to match the type of DeliveryOpaque it needed to be Arc'd.

rust-arroyo/src/backends/kafka/producer.rs
evanh: I don't think we want it to block indefinitely. I'm open to suggestions though.

rust-arroyo/src/backends/kafka/producer.rs
```diff
@@ -118,9 +117,9 @@ impl ArroyoProducer<KafkaPayload> for KafkaProducer {
         // If the producer fails to flush the message out of the buffer, the delivery callback will send an error
         // on the channel.
         // match rx.recv_timeout(Duration::from_secs(5)) {
-        match rx.recv() {
+        match rx.recv_timeout(Duration::from_secs(5)) {
```
untitaker, line 120: for the record i think we should make the produce function here return a future, so to align it with python. but it's not that important and we can postpone it

rust-arroyo/src/backends/kafka/producer.rs
```diff
@@ -98,13 +106,23 @@ impl ArroyoProducer<KafkaPayload> for KafkaProducer {
...
+        // If the producer fails to enqueue the message, this will return an error
         self.producer
             .send(base_record)
             .map_err(|(kafka_error, _record)| ProducerError::from(kafka_error))?;
 
-        Ok(())
+        // If the producer fails to flush the message out of the buffer, the delivery callback will send an error
+        // on the channel.
+        // match rx.recv_timeout(Duration::from_secs(5)) {
+        match rx.recv_timeout(Duration::from_secs(5)) {
```
fpacifici, line 120: Am I missing something or this is making the `produce` call synchronous and make it return only when the message has been sent to Kafka? 
That would not be acceptable. We rely on produce be asynchronous for performance.

fpacifici (changes requested):
Please address my question before merging.

rust-arroyo/src/backends/kafka/producer.rs
fpacifici: Ok, let me see if I understood this correctly:
- The Arroyo producer `produce` step is synchronous here. That does not behave like the rdkafka one.
- But we use the [PRoducer strategy](https://github.com/getsentry/arroyo/blob/main/rust-arroyo/src/processing/strategies/produce.rs#L89) to produce in the consumers and that manages produciton in a separate thread so that would absorb the latency. 

Is that correct?

If yes I think there are still problems:
- If we are not in a consumer arroyo produce becomes synchronous. as the `Producer` strategy can only be used in consumers. This is not ok. 
- In high throughput consumer we can produce thousands of messages in each batch that goes to kafka. 
- - We would keep a channel open for each one of these thousands messages. 
- - I doubt this is not going to be an performance issue. For a produce batch we will have thousands of tasks on the multithread pool waiting for results. This is likely going to create a massive amount of context switches even if the producer releases the thread while waiting (don't know how the wait call is implemented under the hoods).
- uptime checker does not use the Producer strategy https://github.com/getsentry/uptime-checker/blob/main/src/producer/kafka_producer.rs 
- Snuba does not use the producer strategy https://github.com/getsentry/snuba/blob/master/rust_snuba/src/strategies/commit_log.rs#L130. Fair, they run the commit log producer in a separate thread. That would change the behavior nevertheless.

fpacifici: > My understanding was that we wanted this to be synchronous? How else would we track errors when producing properly?

Prodcue should never be synchronous. That's what allows kafka to be fast. Production and consumption are done in batches asynchronously. You fill in a buffer toi produce and rdkafka sends it in the background.
In order to be notified about the result of an operation we use callbacks from rdkafka both in python and rust. That's how we decide when to commit a message in the indexer for example. We commit only when we get a positive confirmation the message was taken by kafka via the callback.

evanh: No I can record the metrics in the callback, that's not hard. I only went through this process because it was my understanding that this behaviour was a bug not a feature. I will close this PR and open a new one.

evanh:
This was fixing the wrong problem. The issue is not that the ArroyoProducer is asynchronous, but that the ProduceStrategy is asynchronous. This change would almost certainly cause performance issues.

Closing this so we can fix the other issue.

PR #476, opened 2025-07-23
WIP: feat(producer): Add an AsyncProducer that returns Future

evanh:
Using this producer allows client code to wait for the future to complete, and capture the resulting
error. Under the hood the FutureProducer creates a oneshot channel that it uses to pass the results
from the produce.


untitaker (approved):
why not use it in the strategy?

evanh:
> why not use it in the strategy?

It's not clear to me that's what we want. If we do that, then anyone using the strategy will wait for their message to be flushed from the buffer, which is different from what is happening now.

PR #477, opened 2025-07-23
feat: Extract timestamp from the message

phacops:
This adds the timestamp associated with the message and set it as part of our `KafkaPayload`.


fpacifici:
Why do we need to copy the timestamp from the Message to the Payload ?

fpacifici:
Timestamp is an attribute of message, which is populated for BrokerMessage as the Kafka message timestamp. you should not need to move it to payload https://github.com/getsentry/arroyo/blob/main/rust-arroyo/src/types/mod.rs#L328-L335

PR #479, opened 2025-08-06
ref: Add more metrics and producer_name tag

untitaker:
ref STREAM-324

See https://github.com/getsentry/relay/pull/5031 for context

### DIFF
--- a/arroyo/backends/kafka/configuration.py
+++ b/arroyo/backends/kafka/configuration.py
@@ -1,4 +1,5 @@
 import copy
+import functools
 import json
 import logging
 from typing import Any, Dict, Mapping, Optional, Sequence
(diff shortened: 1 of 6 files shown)


arroyo/backends/kafka/configuration.py
```diff
@@ -111,10 +199,15 @@ def build_kafka_producer_configuration(
         default_config, bootstrap_servers, override_params
     )
 
+    # Extract client.id to use as producer name for metrics
+    producer_name = broker_config.get("client.id")
```
fpacifici, line 203: Where do we set this ?

untitaker: the idea is that we let the user configure rdkafka's `client.id` (since they are already responsible for coming up with a kafka producer config) and then use that as a tag.

PR #486, opened 2025-10-01
Add option to not reuse subprocesses when using MultiprocessingPool

NicoHinderling:
Python's multiprocessing.pool by default re-uses subprocesses. We have some memory leaks that we'd like to temporarily deal with by creating a new subprocess for every batch that gets processed.

The default functionality will remain the same but this will allow us to set reuse_processes to false and get our desired functionality :)


fpacifici (changes requested):
I don't think you need this 
https://github.com/getsentry/arroyo/blob/main/arroyo/processing/strategies/run_task_with_multiprocessing.py#L530

This is already supported. Just provide a new instance of the ProcessingPool from the StrategyFactory. This will recreate it at every rebalancing

PR #535, opened 2026-04-21
fix(consumer): Don't Crash Rust Consumer on Transport Error

george-sentry:
## Linear
Completes [STREAM-856](https://linear.app/getsentry/issue/STREAM-856/rust-arroyo-consumers-crash-on-transient-errors-during-poll)

## Description
Python and Rust Arroyo consumers handle errors differently during poll.

In Python, consumers raise TransportError if they hit `KafkaError._TRANSPORT` during poll.

https://github.com/getsentry/arroyo/blob/e33f4d6e2612a929c7057b074c9eac6a90f8241e/arroyo/backends/kafka/consumer.py#L501-L502

TransportError is a RecoverableError, so StreamProcessor just swallows the error and goes to the next loop.

https://github.com/getsentry/arroyo/blob/e33f4d6e2612a929c7057b074c9eac6a90f8241e/arroyo/processing/processor.py#L458-L459

Rust doesn't seem to have a concept of recoverable errors, so it just crashes when we hit any error during poll.

https://github.com/getsentry/arroyo/blob/e33f4d6e2612a929c7057b074c9eac6a90f8241e/rust-arroyo/src/processing/mod.rs#L442

We need to update Rust Arroyo to retry / ignore `KafkaError._TRANSPORT`. This PR accomplishes that by checking whether the error is one we consider to be "recoverable", and if it is, we return `Ok(None)`. If it isn't one we consider "recoverable," we return an error as before.

### DIFF
--- a/rust-arroyo/src/backends/kafka/mod.rs
+++ b/rust-arroyo/src/backends/kafka/mod.rs
@@ -81,6 +81,15 @@ impl KafkaConsumerState {
     }
 }
 
+/// Treat `RD_KAFKA_RESP_ERR__TRANSPORT` errors from `librdkafka` like an empty poll the same way Python treats `KafkaError._TRANSPORT`.
+fn kafka_poll_error_is_recoverable_transport(err: &KafkaError) -> bool {
+    matches!(
+        err,
+        KafkaError::MessageConsumption(RDKafkaErrorCode::BrokerTransportFailure)
+            | KafkaError::Global(RDKafkaErrorCode::BrokerTransportFailure)
+    )
+}
+
 fn create_kafka_message(topics: &[Topic], msg: BorrowedMessage) -> BrokerMessage<KafkaPayload> {
     let topic = msg.topic();
     // NOTE: We avoid calling `Topic::new` here, as that uses a lock to intern the `topic` name.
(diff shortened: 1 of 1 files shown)


george-sentry:
My solution is kind of ad hoc, and doesn't really reflect the Python logic. I could do that by extending the Rust `ConsumerError` enum with a new member like `RecoverableError` or `TransportError`.

rust-arroyo/src/backends/kafka/mod.rs
```diff
@@ -81,6 +81,15 @@ impl KafkaConsumerState {
     }
 }
 
+/// Treat `RD_KAFKA_RESP_ERR__TRANSPORT` errors from `librdkafka` like an empty poll the same way Python treats `KafkaError._TRANSPORT`.
+fn kafka_poll_error_is_recoverable_transport(err: &KafkaError) -> bool {
```
untitaker, line 85: maybe a bad idea, but you can make this a macro like such:

```
macro_rules recoverable_error_pattern {
    () => ((KafkaError::MessageConsumption(RDKafkaErrorCode::BrokerTransportFailure)
            | KafkaError::Global(RDKafkaErrorCode::BrokerTransportFailure));
}

...

match res {
    Some(Err(recoverable_error_pattern!())) => ...,
    Some(Err(err)) => ...
}
```

then the exhaustiveness and dead code checks of the compiler will continue to be effective because you avoided one `if matches!()`

untitaker (approved):

PR #536, opened 2026-04-21
feat(run_task): Add better_backpressure flag to avoid re-running function

untitaker:
## Summary

- Adds `better_backpressure=True` flag to `RunTask` strategy
- When enabled, stores transformed messages on backpressure and retries in `poll()` instead of re-raising `MessageRejected`
- Prevents user function from running multiple times on the same message during backpressure

## Context

The span buffer's flusher sometimes applies backpressure, causing increased CPU load on Redis. We suspect this is because `add-buffer.lua` is being executed multiple times per message when the current `RunTask` implementation re-runs the function on each retry.

With this fix (opt-in via `better_backpressure=True`), the function runs exactly once per message, matching the Rust arroyo behavior.

Ref STREAM-881

## Test plan

- [x] Existing tests pass
- [x] mypy clean
- [ ] Enable in span buffer consumer and monitor Redis CPU

### DIFF
--- a/arroyo/processing/strategies/run_task.py
+++ b/arroyo/processing/strategies/run_task.py
@@ -48,7 +48,6 @@ class RunTask(
         self.__next_step = next_step
         self.__better_backpressure = better_backpressure
         self.__message_carried_over: Optional[Message[TResult]] = None
-        self.__closed = False
 
     def submit(
         self, message: Message[Union[FilteredPayload, TStrategyPayload]]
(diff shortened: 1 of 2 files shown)


arroyo/processing/strategies/run_task.py
```diff
@@ -24,12 +25,13 @@ def __new__(
         cls,
         function: Callable[[Message[TStrategyPayload]], TResult],
         next_step: ProcessingStrategy[Union[FilteredPayload, TResult]],
+        better_backpressure: bool = False,
```
markstory, line 28: If this works well it seems like it could be the default going forward.

markstory (approved):

arroyo/processing/strategies/run_task.py
untitaker: i think this should be the intent, just didn't want to take the risk now

PR #545, opened 2026-06-28
feat(metrics): add distribution metric support

phacops:
## Summary

Adds support for **distribution** metrics to Arroyo's metrics backend interface, alongside the existing `increment`, `gauge`, and `timing` methods.

Distribution metrics let backends record the full distribution of a set of values (e.g. percentiles), which is useful for measuring things like batch sizes or latencies where aggregate statistics matter.

## Changes

- **`arroyo/utils/metrics.py`**: Add the abstract `distribution` method to the `Metrics` protocol and a no-op implementation in `DummyMetricsBackend`.
- **`tests/metrics.py`**: Add a `Distribution` call type and implement `distribution` in the testing backend so recorded distribution calls can be asserted in tests.
- **`tests/utils/test_metrics.py`**: Add a test covering the testing backend's `distribution` method.
- **`docs/source/metrics.rst`**: Document `distribution` in the example backend implementation.
- **`CHANGELOG.md`**: Add an Unreleased entry.

## Notes

This is purely additive to the metrics interface. Backends implementing the `Metrics` protocol will now need to provide a `distribution` method.

## Testing

- `pytest tests/utils/test_metrics.py` passes.
- Verified `DummyMetricsBackend` and the testing backend still satisfy the `runtime_checkable` `Metrics` protocol.

https://claude.ai/code/session_019ASHzHqVirSvv3vBRYaQPS

---
_Generated by [Claude Code](https://claude.ai/code/session_019ASHzHqVirSvv3vBRYaQPS)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog and versioning
 
+## Unreleased
+
+### New Features ✨
+
+- Add `distribution` method to the metrics backend interface for recording distribution metrics
+
 ## 2.40.3
 
 ### New Features ✨
--- a/arroyo/utils/metrics.py
+++ b/arroyo/utils/metrics.py
@@ -44,6 +44,15 @@ class Metrics(Protocol):
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def distribution(
+        self, name: MetricName, value: Union[int, float], tags: Optional[Tags] = None
+    ) -> None:
+        """
+        Records a distribution metric.
+        """
+        raise NotImplementedError
+
 
 class DummyMetricsBackend(Metrics):
     """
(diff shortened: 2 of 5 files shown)


untitaker:
> Distribution metrics let backends record the full distribution of a set of values (e.g. percentiles), which is useful for measuring things like batch sizes or latencies where aggregate statistics matter.

okay but we already use timing everywhere, and arroyo isn't prescriptive about how you aggregate those values already. so you could impl `timing` as a histogram, sampled quantiles, or both, or anything else. is the plan to introduce new metrics that use `distribution`? right now this dead code.

(don't have access to the claude session so I can't figure out what the bigger picture here is)

phacops:
With us moving away from `veneur` and using `ddog-statsd`, we can now use real distribution metrics instead of the histogram + split into different percentile metrics at `veneur`.<br><br>This PR is adding support for these metrics in Arroyo so we can use them in Snuba consumers [https://github.com/getsentry/snuba/blob/10cd401703ce9bd5c838ad38a99a834d4c963957/rust_snuba/src/metrics/statsd.rs#L124-L162](<https://github.com/getsentry/snuba/blob/10cd401703ce9bd5c838ad38a99a834d4c963957/rust_snuba/src/metrics/statsd.rs#L124-L162>)

onewland (approved):

PR #547, opened 2026-07-03
feat(producer): add FutureTrackingProducer

bmckerry:
This PR adds `FutureTrackingProducer`, which is wrapper for a producer object that can track futures internally, then pass them into user code for processing when `FutureTrackingProducer.collect_futures()` is called.

Functionally, this is a copy & paste of [TaskProducer](https://github.com/getsentry/taskbroker/blob/main/clients/python/src/taskbroker_client/worker/producer.py#L24) in the taskbroker repo (with the addition of tracking futures optionally based off of an environment variable). I plan to fully replace `TaskProducer` with this producer in order to have a single producer we can use for tasks/consumers/etc in Sentry (and elsewhere).

### DIFF
--- a/arroyo/backends/kafka/producer.py
+++ b/arroyo/backends/kafka/producer.py
@@ -1,0 +1,126 @@
+import atexit
+import os
+from collections import defaultdict, deque
+from collections.abc import Callable
+from concurrent.futures import Future
+from typing import Any, Protocol, Sequence
+
+from arroyo.backends.abstract import ProducerFuture, SimpleProducerFuture
+from arroyo.backends.kafka import KafkaPayload
+from arroyo.types import BrokerValue, Partition, Topic
+
+TASK_PRODUCER_MAX_PENDING_FUTURES = 10_000
+
+_pending_futures: defaultdict[
+    str, deque[ProducerFuture[BrokerValue[KafkaPayload]]]
+] = defaultdict(lambda: deque(maxlen=TASK_PRODUCER_MAX_PENDING_FUTURES))
+"""
+_pending_futures is global as a process may have several `FutureTrackingProducer` instances,
+and we want to collect futures from all running instances with a single call.
+Keys are the names of each `FutureTrackingProducer` instance in the current process, values are
+deques with a maxlen to prevent unbounded queue size in case we're set to track futures,
+but `collect_futures()` is never called.
+"""
+
+
+class CloseableProducerProtocol(Protocol):
+    """Interface used by FutureTrackingProducer. Represents a producer that has a shutdown method."""
+
+    def produce(
+        self, dest: Topic | Partition, payload: KafkaPayload
+    ) -> ProducerFuture[BrokerValue[KafkaPayload]]:
+        ...
+
+    def close(self) -> Future[None]:
+        ...
+
+
+class FutureTrackingProducer:
+    """
+    FutureTrackingProducer is a producer abstraction that optionally registers futures
+    within the `_pending_futures` dict (depending on if the `ARROYO_TRACK_PRODUCER_FUTURES`
+    env var is set). Applications can then call the `collect_futures()` static method to
+    get all registered futures, await them, etc.
+
+    Args:
+        name: Unique identifying name of this FutureTrackingProducer. Used to key `_pending_futures`.
+        producer_factory: Callable that returns a producer object.
+    """
+
+    def _should_track_futures(self) -> bool:
+        return os.environ.get("ARROYO_TRACK_PRODUCER_FUTURES", "").lower() == "true"
+
+    def __init__(
+        self,
+        name: str,
+        producer_factory: Callable[[], CloseableProducerProtocol],
+    ) -> None:
+        self.name = name
+        self._producer_factory = producer_factory
+        self._inner_producer: CloseableProducerProtocol | None = None
+        self._track_futures = self._should_track_futures()
+
+    def _get(self) -> CloseableProducerProtocol:
+        if self._inner_producer is None:
+            self._inner_producer = self._producer_factory()
+            atexit.register(self._shutdown)
+        return self._inner_producer
+
+    def track_future(self, future: ProducerFuture[BrokerValue[KafkaPayload]]) -> None:
+        _pending_futures[self.name].append(future)
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


arroyo/backends/kafka/producer.py
```diff
@@ -0,0 +1,126 @@
...
+        _pending_futures[self.name].append(future)
+
+    @staticmethod
+    def collect_futures() -> dict[str, set[ProducerFuture[BrokerValue[KafkaPayload]]]]:
+        """
+        Clears the `_pending_futures` dict, and returns a copy with all values converted to sets.
+        """
+        pending_copy = _pending_futures.copy()
+        _pending_futures.clear()
+        return {name: set(val) for name, val in pending_copy.items()}
```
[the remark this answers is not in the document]
bmckerry: This is how the existing TaskProducer works, because the same thread that calls `produce` is always the thread that calls `.collect_futures()`. I don't expect anyone to call `collect_futures()` outside of that context.

arroyo/backends/kafka/producer.py
```diff
@@ -0,0 +1,129 @@
...
+            if not isinstance(future, SimpleProducerFuture):
+                for c in callbacks:
+                    future.add_done_callback(c)
+            else:
+                raise RuntimeError(
+                    (
+                        "Cannot add callbacks to SimpleProducerFuture, either remove the callbacks "
+                        "or instantiate your producer with `use_simple_futures=False`."
+                    )
+                )
```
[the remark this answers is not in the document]
bmckerry: We need to get the future from producing in order to see if its a `SimpleProducerFuture`, unfortunately.

arroyo/backends/kafka/producer.py
```diff
@@ -0,0 +1,129 @@
...
+        self.metrics = (
+            metrics_backend if metrics_backend is not None else DummyMetricsBackend()
+        )
+        self._track_futures = self._should_track_futures()
+
+    def _get(self) -> CloseableProducerProtocol:
+        if self._inner_producer is None:
+            self._inner_producer = self._producer_factory()
+            atexit.register(self._shutdown)
+        return self._inner_producer
```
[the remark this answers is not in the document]
bmckerry: While possible I feel like this isn't a significant issue (and I'm not sure how to guard against this).

arroyo/backends/kafka/producer.py
```diff
@@ -0,0 +1,126 @@
...
+from collections import defaultdict, deque
+from collections.abc import Callable
+from concurrent.futures import Future
+from typing import Any, Protocol, Sequence
+
+from arroyo.backends.abstract import ProducerFuture, SimpleProducerFuture
+from arroyo.backends.kafka import KafkaPayload
+from arroyo.types import BrokerValue, Partition, Topic
+
+TASK_PRODUCER_MAX_PENDING_FUTURES = 10_000
```
markstory:
```suggestion
PRODUCER_MAX_PENDING_FUTURES = 10_000
```

Does the task prefix still make sense with this being in arroyo?

arroyo/backends/kafka/producer.py
```diff
@@ -0,0 +1,126 @@
...
+
+_pending_futures: defaultdict[
+    str, deque[ProducerFuture[BrokerValue[KafkaPayload]]]
+] = defaultdict(lambda: deque(maxlen=TASK_PRODUCER_MAX_PENDING_FUTURES))
```
markstory:
```suggestion
] = defaultdict(lambda: deque(maxlen=PRODUCER_MAX_PENDING_FUTURES))
```

arroyo/backends/kafka/producer.py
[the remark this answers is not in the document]
markstory: You can guard against it with a [threading.Lock](https://docs.python.org/3/library/threading.html#lock-objects).

You could do something like

```python
    def __init__(
        self,
        name: str,
        producer_factory: Callable[[], CloseableProducerProtocol],
    ) -> None:
        self.name = name
        self._producer_factory = producer_factory
        self._inner_producer: CloseableProducerProtocol | None = None
        self._track_futures = self._should_track_futures()
        self._inner_lock = threading.Lock()

    def _get(self) -> CloseableProducerProtocol:
        if self._inner_producer is None:
            with self._inner_lock
                self._inner_producer = self._producer_factory()
                atexit.register(self._shutdown)
        return self._inner_producer
```

edit: updated the order of none + lock checks to avoid contention.

arroyo/backends/kafka/producer.py
```diff
@@ -0,0 +1,126 @@
...
+    @staticmethod
+    def collect_futures() -> dict[str, set[ProducerFuture[BrokerValue[KafkaPayload]]]]:
+        """
+        Clears the `_pending_futures` dict, and returns a copy with all values converted to sets.
+        This generally should not be called by users, and should only be called if you want to
+        do something based on the outcome of a _group_ of futures, and not just the outcomes of individual futures.
+        For the latter, use delivery callbacks.
+
+        This operation should also always be called in the same thread as the one calling `produce()`, since there's
+        no locking around `_pending_futures`.
```
markstory: Because there isn't any IO/native calls in this method, the GIL gives you a soft-lock around this. This could be made threadsafe with another `threading.Lock`.

enochtangg (approved):

tests/backends/test_future_tracking_producer.py
```diff
@@ -0,0 +1,127 @@
...
+
+def get_dummy_producer(use_simple_futures: bool) -> DummyProducer:
+    return DummyProducer(use_simple_futures=use_simple_futures)
+
+
+@pytest.fixture(autouse=True)
+def clear_pending_futures() -> Iterator[None]:
+    _pending_futures.clear()
+    yield
+    _pending_futures.clear()
```
markstory, line 56: Should `FutureTrackingProducer` have a static method to do this?

markstory (approved):

tests/backends/test_future_tracking_producer.py
bmckerry: Going to hold off on this since we'd only use this in tests IMO, calling `collect_futures()` should be the way to clear `_pending_futures` when running this in a regular environment.

PR #551, opened 2026-07-13
feat(producer): add metrics for poll frequency

bmckerry:
Adds a sampled counter metric that's incremented when the arroyo `KafkaProducer` worker thread polls the internal librdkafka producer. Arroyo's metrics backend doesn't expose a `sample_rate` option so I did this manually by incrementing every X polls by X (configurable). Metric tracking is enabled when the produce is instantiated, I plan on enabling in Sentry thru an option.

This is being added in an attempt to root cause INC-2308.

### DIFF
--- a/arroyo/backends/kafka/consumer.py
+++ b/arroyo/backends/kafka/consumer.py
@@ -758,11 +758,17 @@ class KafkaConsumer(Consumer[KafkaPayload]):
 
 class KafkaProducer(Producer[KafkaPayload]):
     def __init__(
-        self, configuration: Mapping[str, Any], use_simple_futures: bool = False
+        self,
+        configuration: Mapping[str, Any],
+        use_simple_futures: bool = False,
+        record_poll_metrics: bool = False,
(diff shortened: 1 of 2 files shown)


arroyo/backends/kafka/consumer.py
```diff
@@ -779,8 +785,22 @@ def __worker(self) -> None:
         after a shutdown request has been issued (via ``close``) and all
         in-flight messages have been delivered.
         """
+        POLL_METRIC_FREQUENCY = 10
+        poll_count = 0
```
george-sentry: Does this ever reset? I imagine it could get quite large after running for a long time.

george-sentry: Also, would it make sense to define `POLL_METRIC_FREQUENCY` elsewhere since it seems to be a constant? Or rename to be lowercase like a normal variable? Or make it configurable at runtime?

george-sentry (approved):

PR #553, opened 2026-07-22
ref(producer): Track futures based on arg, not env var

bmckerry:
This PR reworks `FutureTrackingProducer` to track futures based on args passed to the producer, not an environment variable. After release I'll add a helper function to Sentry that will generate an instance of FTP with this arg set depending on if we're in a taskworker or not.


untitaker (approved):
doesn't look like a CLI to me, but i approve of the diff

PR #556, opened 2026-07-23
feat(producer): Add a produce_with_callback function to the producer

evanh:
Previously users didn't have a way of handling the results from the AsyncProducer. Add a way for a
callback to be passed in that gets executed when the produce finishes. That callback will receive
the result of the produce (including any errors).


untitaker (approved):
what is the context where this is needed?

evanh:
> what is the context where this is needed?

https://linear.app/getsentry/issue/STREAM-311/arroyo-kafkaproducer-should-allow-the-client-to-react-when-the-message

Just working through our backlog.

untitaker:
ok, that ticket also has no context for _why_, but it says this:

> We can either expose the callback to be customizable so the client can react to the successful delivery or make the producer return a future like in python.

so we already return a future from produce() -- maybe we didn't back then?

evanh:
> so we already return a future from produce() -- maybe we didn't back then?

Fair point. OK well I'll close this then.

PR #560, opened 2026-08-19
feat(runtime): Implement pull based arroyo

tryangul:
### What
Pull-based (async stream) processing runtime — an alternative to the push-based `ProcessingStrategy` model. Gated behind Cargo level `pull` feature flag.

### How
New `processing/stream/` module. Fully additive — no existing push code changed (except a couple visibility mods).

**Core:**
- `Stage` trait — async `process(&self, envelope) -> StageResult`
- `StageResult` — `Emit`, `Drop`, `Skip`, `Reject`, `Fail`, `Exit`
- `PipelineEnvelope<T>` — payload + metadata + raw bytes
- `PipelineExt` combinators — `.apply()`, `.apply_concurrent()`, `.apply_with_timer()`, `.on_next()`, `.on_reject()`, `.commit()`, `.run()`
- `Pipeline` trait — `stream()` accepts a source stream, returns a processed stream

**Batching:**
- `Buffer<T>` trait with associated `Output` type
- `BatchStage` — row count + byte size triggers
- `FlushableStage` trait for time-based flushing
- `FlushTimer` — coarsetime watermarks + interval tick, injectable `Clock`

**Sources + lifecycle:**
- `PullSource` trait (object-safe) — `KafkaSource` with `StreamConsumer`, rebalance detection, graceful shutdown
- `PipelineRunner` — rebalance restart loop (might need refactoring but the loop is correct)
- `OffsetTracker` / `OffsetCollector` — time-throttled offset commits

**Handlers:**
- `RejectionHandler` trait with `on_reject` combinator allow for "railroad oriented" design where rejected items flow directly here. `DlqHandler`, `LogHandler` implementations
- `NextHandler` trait allows side effects without mutation. `KafkaProducerHandler` as an example.

**Tests:** 71 passing

### Notes
- Rebalance limitation: `StreamConsumer` callback runs inline during poll. Cannot block to drain. Best-effort offset commit after unassign. Enabling proper drain requires  using `BaseConsumer` directly.
- `FlushTimer` DI needs redesign — timer interval should be injectable instead of created by combinator. Deferred.


rust-arroyo/src/processing/stream/offset_tracker.rs
```diff
@@ -0,0 +1,93 @@
...
+            last_commit_time: coarsetime::Instant::recent(),
+            last_record_time: coarsetime::Instant::recent(),
+            commit_frequency: commit_frequency.into(),
+        }
+    }
+
+    /// Record the offset for a partition.
+    pub fn track(&mut self, partition: Partition, offset: u64) {
+        self.offsets.insert(partition, offset);
+    }
```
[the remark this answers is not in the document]
tryangul: We use `buffered` so everything is processed in order. I don't believe this is valid.

rust-arroyo/src/processing/stream/ext.rs
```diff
@@ -0,0 +1,305 @@
...
+                StageResult::Reject { metadata, .. } => {
+                    tracker.track(metadata.partition, metadata.offset + 1);
+                }
+                StageResult::Fail(err) => {
+                    let _ = tracker.flush();
+                    return Err(err);
+                }
+                StageResult::Exit(reason) => {
+                    tracker.flush()?;
+                    return Ok(reason);
```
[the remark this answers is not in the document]
tryangul: We could swallow the error, but I have not actually seen this fail. Would want to test before swallowing.

rust-arroyo/src/processing/stream/batch/stage.rs
```diff
@@ -0,0 +1,403 @@
...
+            metadata.offset = max_offset;
+        }
+        self.offsets.clear();
+
+        self.row_trigger = SizeTrigger::new(max_rows);
+        self.byte_trigger = SizeTrigger::new(max_bytes);
+
+        Some(StageResult::Emit(PipelineEnvelope::new(
+            items, metadata, raw,
+        )))
```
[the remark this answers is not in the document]
tryangul: Dupe of https://github.com/getsentry/arroyo/pull/560#discussion_r4030344231

rust-arroyo/src/processing/stream/batch/stage.rs
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        let items = self.buffer.flush();
+        let mut metadata = self.last_metadata.take()?;
+        let raw = self.last_raw.take()?;
+
+        if let Some(&max_offset) = self.offsets.get(&metadata.partition) {
+            metadata.offset = max_offset;
+        }
+        self.offsets.clear();
```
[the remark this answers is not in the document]
tryangul: Worst case is some replay, but it does look like a map here would be optimal.

rust-arroyo/src/processing/stream/offset_tracker.rs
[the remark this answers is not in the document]
tryangul: Actually, because the rejects can fast path, this does present an issue.

tryangul:
The cursor feedback looks valid. I think the fixes are roughly a day of focused work. I still lean towards getting this in as is and fast-follow fixing things, as the pull model isn't actually used any where. The question is who will review it.

tryangul:
Updated to put this API behind a Cargo feature flag.

rust-arroyo/src/processing/stream/offset_tracker.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        if !force && coarsetime::Instant::recent() - self.last_commit_time <= self.commit_frequency
+        {
+            return Ok(());
+        }
+
+        self.committer.commit_offsets(&self.offsets)?;
+        self.offsets.clear();
+        self.last_commit_time = coarsetime::Instant::recent();
+        Ok(())
```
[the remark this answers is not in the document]
tryangul: Same behavior as normal arroyo afiact

rust-arroyo/src/processing/stream/offset_tracker.rs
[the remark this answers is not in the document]
tryangul: After looking into this, to get parity with push, we just need to stop tracking offsets for Drop / Reject. For things without batching it'd be nice to keep track of them, but that's not necessary for parity.

~2 line deletion + remove tests.

rust-arroyo/src/processing/stream/ext.rs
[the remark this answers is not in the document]
tryangul: I believe this also exists in the push model. Not a regression. Closing.

rust-arroyo/src/processing/stream/collector/stream_collector.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Receives pipeline events from the `run()` terminal combinator.
+///
+/// Implement this to control what happens when items are emitted,
+/// dropped, or rejected. `OffsetCollector` tracks offsets for Kafka
+/// commit. `NoopCollector` drains without side-effects.
+pub trait StreamCollector<T>: Send + Sync {
+    fn on_emit(&mut self, envelope: &PipelineEnvelope<T>);
+    fn on_drop(&mut self, metadata: &MessageMetadata);
+    fn on_reject(&mut self, metadata: &MessageMetadata);
+    fn on_complete(&mut self) -> Result<(), BoxError>;
```
fpacifici, line 13: What is this meant to do ?

rust-arroyo/src/processing/stream/pipeline_runner.rs
```diff
@@ -0,0 +1,248 @@
...
+            let result = pipeline
+                .stream(source.stream())
+                .commit(&mut tracker)
+                .await?;
+
+            match result {
+                PipelineExit::Rebalance => {
+                    tracing::info!("Rebalance detected, restarting pipeline...");
+                }
+                PipelineExit::Shutdown | PipelineExit::Complete => return Ok(()),
```
fpacifici, line 57: Fyi, 
when this will be ready for prod, the observability here has to be different. We need at least:
- Logs for all partitions revoked and assigned. 
- metrics to count revocations and assignment.

rust-arroyo/src/processing/stream/pipeline_envelope.rs
```diff
@@ -0,0 +1,88 @@
...
+        let partition = Partition::new(topic, msg.partition() as u16);
+        let time_millis = msg.timestamp().to_millis().unwrap_or(0);
+        let timestamp =
+            DateTime::from_timestamp_millis(time_millis).unwrap_or(DateTime::<Utc>::MIN_UTC);
+
+        let kafka_payload = KafkaPayload::new(
+            msg.key().map(|k| k.to_vec()),
+            msg.headers().map(|h| h.into()),
+            msg.payload().map(|p| p.to_vec()),
+        );
```
fpacifici, line 74: If I get this right, this would induce a payload copy. 
Is there any way to Move around the PipelineEnvelope without copying the payload. This is problematic on high throughput pipelines like spans

rust-arroyo/src/processing/stream/pipeline_envelope.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        let metadata = MessageMetadata {
+            partition,
+            offset: msg.offset() as u64,
+            timestamp,
+        };
+
+        Self {
+            payload: kafka_payload.clone(),
```
fpacifici, line 83: See above, this is a second copy.

rust-arroyo/src/processing/stream/pipeline_envelope.rs
```diff
@@ -0,0 +1,88 @@
...
+///
+/// Three concerns, three fields:
+///   - `payload` — the current transformed data (changes at each stage)
+///   - `metadata` — partition/offset/timestamp for commit tracking
+///   - `raw` — original Kafka bytes for DLQ
+pub struct PipelineEnvelope<T> {
+    pub payload: T,
+    pub metadata: MessageMetadata,
+    pub raw: KafkaPayload,
+}
```
fpacifici, line 26: IT seems to me we are loosing the headers here.
Please preserve them. We use them to do fast filtering without payload parsing or copy in many pipelines.

rust-arroyo/src/processing/stream/source.rs
```diff
@@ -0,0 +1,212 @@
...
+use super::BoxError;
+
+/// Trait for pipeline sources. Provides a stream of raw Kafka payloads,
+/// an offset committer, and graceful shutdown.
+///
+/// Object-safe — can be used as `Box<dyn PullSource>` or `Arc<dyn PullSource>`.
+pub trait PullSource: Send + Sync {
+    /// Returns a stream of Kafka messages. Ends on rebalance, shutdown,
+    /// or when the source is exhausted. Callable multiple times — each
+    /// call returns a fresh stream for the current partition assignment.
```
fpacifici, line 31: What is the goal of calling this multiple times ?

rust-arroyo/src/processing/stream/source.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+/// ConsumerContext that signals the pipeline on partition revocation.
+///
+/// NOTE: The rebalance callback runs inline during StreamConsumer::poll_next
+/// (same thread/task as the async stream). We cannot block here to drain
+/// the pipeline — that would deadlock. Instead, we just notify and return.
+/// rdkafka proceeds with unassign immediately. The pipeline drains after
+/// unassign, so offset commits are best-effort.
+///
```
fpacifici, line 48: I think this may be a problem.
If the partitions are revoked immediately (the revocation callback ends without flushing), I think, the commit will always fail. 
So there are two problems:
- The default contract we provide today is that on rebalance the steps will be flushed and committed unless errors happen. This would change that contract. It may be ok, but it can have application implications. Arroyo always bent over backwards to ensure we would achieve exactly once in most cases as many application features do not behave well with duplicate data. If we move to a place where data is reprocessed at each rebalance we will break something.
- The commit here always fails as the revocation callback seems to complete before we attempt to commit.

rust-arroyo/src/processing/stream/source.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn stream(&self) -> BoxStream<'_, StageResult<KafkaPayload>> {
+        let shutdown = self.shutdown.clone();
+        let revoke = &self.revoke;
+
+        Box::pin(stream! {
+            let mut kafka_stream = self.consumer.stream();
+
+            loop {
+                tokio::select! {
+                    msg = kafka_stream.next() => {
```
fpacifici, line 152: One general note on observability.
We rely a lot on three liveness metrics to manage consumers. These should be propagated into this implementation whether in this PR or not.
1. Poll time. This is the  inverse of our utilization metric. The longer we wait to get the next Kafka metrics the more free time the consumer has. This or a replacement is critical. Now since here we have async code in the way it may be hard to record that. Though a consumer that is not overloaded should still be waiting here.
2. Execution breakdown. https://app.datadoghq.com/s/FH6-Y3/ung-vsh-wsz poll vs processing vs callbacks.
3. Run count: the number of iterations of this loop. If the run count gets lower it means we are stuck somewhere. With async you can even do it in a more elegant way: have a task that sleep 1 second and then record a metric. If that metric is recorded less than once per second it means the main thread is busy or stuck.

rust-arroyo/src/processing/stream/source.rs
```diff
@@ -0,0 +1,212 @@
...
+                partition.index as i32,
+                rdkafka::Offset::Offset(*offset as i64),
+            )
+            .map_err(|e| Box::new(e) as BoxError)?;
+        }
+        // Async commit matches the existing push model's behavior. The broker
+        // may not ack before we clear tracked offsets, but this is acceptable —
+        // worst case on crash is re-processing already-committed messages.
+        self.consumer
+            .commit(&tpl, CommitMode::Async)
```
fpacifici, line 209: How does a commit error turn into a crash ? 
Also see the issue on re-processing data https://github.com/getsentry/arroyo/pull/560/changes#r4031962471

rust-arroyo/src/processing/stream/stage.rs
```diff
@@ -0,0 +1,130 @@
...
+    Ignored,
+}
+
+/// A processing stage in a pull-based pipeline.
+///
+/// Unified trait for all stage types:
+///   - 1:1 transforms — return Emit(envelope)
+///   - Filters — return Drop(metadata) to filter with offset tracking
+///   - Batching — return Skip while accumulating, Emit when flushing
+///   - Errors — return Reject or Fail
```
fpacifici, line 105: In streams we are adding routers and broadcasts 
https://github.com/getsentry/streams/blob/main/sentry_streams/sentry_streams/examples/broadcast.py.

Do you have a plan on how to support branches processing offsets at different speeds ?
In streams we introduced the concept of watermark that propagates through all branches and are tracked by the collector to decide when to commit.

Cargo.toml
```diff
@@ -27,12 +28,15 @@ serde = { version = "1.0.137", features = ["derive"] }
 serde_json = "1.0.81"
 thiserror = "1.0"
 tokio = { version = "1.19.2", features = ["full"] }
+futures = "0.3"
+tokio-stream = { version = "0.1", features = ["time"] }
+tokio-util = "0.7"
```
fpacifici, line 33: Please sync with @lvthanh03  on the dependencies you are adding as he is trying to embed arroyo into several applications and he will have to keep the dependencies under control.

rust-arroyo/src/processing/stream/offset_tracker.rs
```diff
@@ -0,0 +1,93 @@
...
+            offsets: Default::default(),
+            last_commit_time: coarsetime::Instant::recent(),
+            last_record_time: coarsetime::Instant::recent(),
+            commit_frequency: commit_frequency.into(),
+        }
+    }
+
+    /// Record the offset for a partition.
+    pub fn track(&mut self, partition: Partition, offset: u64) {
+        self.offsets.insert(partition, offset);
```
fpacifici, line 53: What if offsets come out of order ?
Today arroyo standard strategies emit in order whether or not the messages are processed in order.
The user can build out of order processing by providing their own commit policy that tracks offsets the way they want. 
If this is not supported, it has to be documented. Also I'd recommend thinking about how to build pipelines with branches where tracking offsets in order is harder.

rust-arroyo/src/processing/stream/offset_tracker.rs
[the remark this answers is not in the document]
fpacifici: > to stop tracking offsets for Drop / Reject.

Would we still commit in a consumer that may be dropping all the messages for a period of time ? 
That can happen

rust-arroyo/src/processing/stream/ext.rs
```diff
@@ -0,0 +1,305 @@
...
+
+/// Run a stage on an envelope and record metrics.
+async fn run_stage<S: Stage>(stage: &S, envelope: PipelineEnvelope<S::In>) -> StageResult<S::Out> {
+    let start = coarsetime::Instant::now();
+    let result = stage.process(envelope).await;
+
+    timer!("arroyo.stage.duration", start.elapsed(), "stage" => stage.name());
+    match &result {
+        StageResult::Emit(_) => {
+            counter!("arroyo.stage.success", 1, "stage" => stage.name());
```
fpacifici, line 26: Is this counter buffered ? We cannot afford to produce metrics at every step. 
Thanks for adding the stage name.

rust-arroyo/src/processing/stream/stage.rs
```diff
@@ -0,0 +1,130 @@
...
+/// The framework handles error routing (DLQ), metrics, and offset tracking.
+/// The process method is async to support stages that do I/O.
+pub trait Stage: Send + Sync {
+    type In: Send;
+    type Out: Send;
+
+    fn process(
+        &self,
+        envelope: PipelineEnvelope<Self::In>,
+    ) -> impl Future<Output = StageResult<Self::Out>> + Send;
```
fpacifici, line 116: Have you considered making this async? If not why ?

rust-arroyo/src/processing/stream/ext.rs
```diff
@@ -0,0 +1,305 @@
...
+            while let Some(item) = upstream.next().await {
+                yield item.map_emit(|e| run_stage(&stage, e)).await;
+            }
+        }
+    }
+
+    /// Apply a processing stage concurrently to up to `concurrency` Emit
+    /// envelopes at once. Results are yielded in input order.
+    /// Non-Emit items pass through immediately.
+    fn apply_concurrent<S>(
```
fpacifici, line 71: One high level remark around concurrency and parallelism.
We generally have two types of scenarios:
- CPU intensive parallelism
- IO intensive consumers.
It seems to me you would use `apply` for processing on the main thread asking the user to ensure they do not take over the main thread for too long (already the case in arroyo push mode).
THis would replace the RunInThread step for IO workload. 

Two questions:
- How do you envision parallelism for CPU intensive workload will work ? Is it up to the stage to manage a thread pool ? In that case what would the stage return ?
- Have you considered running all apply operations concurrently out of the main thread to guarantee we will always poll before getting kicked out ?

rust-arroyo/src/processing/stream/ext.rs
```diff
@@ -0,0 +1,305 @@
...
+    where
+        S: Stage<In = T>,
+    {
+        assert!(concurrency > 0, "concurrency must be at least 1");
+        let stage = Arc::new(stage);
+        self.map(move |item| {
+            let stage = Arc::clone(&stage);
+            async move { item.map_emit(|e| run_stage(&*stage, e)).await }
+        })
+        .buffered(concurrency)
```
fpacifici, line 85: How would we manage a scenario where the consumer is in a backpressure state for longer than the max poll time ? 
Imagine this scenario: We try to write on Clickhouse via apply_concurrent, which means we can still run the main thread and poll from kafka. This operation takes longer than poll time.
In the push model we pause the consumer after a certain time in backpressure mode, so we can keep polling from kafka without receiving messages we would have to accumulate. This prevents kafka from kicking the consumer out.
This is a use case to support.

rust-arroyo/src/processing/stream/handlers/kafka_producer.rs
```diff
@@ -0,0 +1,32 @@
...
+            topic,
+        }
+    }
+}
+
+impl NextHandler<KafkaPayload> for KafkaProducerHandler {
+    async fn handle(&self, envelope: &PipelineEnvelope<KafkaPayload>) -> Result<(), BoxError> {
+        self.producer
+            .produce(&self.topic, envelope.payload.clone())
+            .map_err(|e| Box::new(e) as BoxError)
```
fpacifici, line 30: Does anybody listens to the callback from the producer to ensure we do not commit before we have certainty the message was produced ?

fpacifici (commented):
Thanks.
I think this direction is promising. Though there are areas to work on, Please see the comment in line.

Specifically areas I think are improving the current state:
- async support that can keep the main thread clear and unlocks better concurrency patterns. Though this means the application needs to run on an async runtime.
- Simplification of the consumer. I think it should be possible to support rebalancing and commit callbacks without making it too complex 
- Better abstractions for the application 
- The Result is a better abstraction for DLQ than the exception
- I like offset management better.
- It fits much better than the push arroyo as a runtime for streams https://github.com/getsentry/streams

I would recommend getting a review from other people in the streaming team and define a path  forward on this.

rust-arroyo/src/processing/stream/stage.rs
tryangul: `async fn` desugars to a function that returns `impl Future`. I have to use this form in order to and `Send` bound to enable concurrency.

rust-arroyo/src/processing/stream/ext.rs
tryangul: Our DD client (I believe `DogStatsD`) does buffer and flush at an interval so our network calls should be batched. Also, the metrics and tags are all borrows so we should avoid heap allocations. 

My tack here is to run it and only add more explicit buffering / batching if we see it. Fwiw, I didn't see any performance problems running this for eap items in s4s2.

rust-arroyo/src/processing/stream/pipeline_envelope.rs
tryangul: Good catch. I don't remember coding anything specifically around headers, so it's probably a gap.

rust-arroyo/src/processing/stream/pipeline_envelope.rs
tryangul: This one I believe we have to do to copy out of the rdkafka buffer. There is definitely more copying going on, which I believe you called out. Will address there.

rust-arroyo/src/processing/stream/pipeline_envelope.rs
tryangul: Good catch. I think we don't need to clone here. I'll try removing this. But we do need to clone for the raw payload below.

If we drop this first clone, I believe we are roughly equivalent push. Pull clones in raw and Push clones in `buffered` in `dlq.rs`

rust-arroyo/src/processing/stream/pipeline_runner.rs
tryangul: Gap noted. Great call out.

rust-arroyo/src/processing/stream/source.rs
tryangul: My goal is to have the Source live across rebalances. So we stay connected to the broker and just build a new stream with the new partitions that we can pass to a freshly created pipeline ensuring no stale state gets shared.

rust-arroyo/src/processing/stream/source.rs
tryangul: Yea, this was a vexing one. I agree we need it for a true general streaming library. However, it didn't present problems with my Snuba POC, so I de-prioritized it.

*The nitty gritty*:
I wasn't able to implement a proper drain countdown latch because the streaming rdkafka client runs the poll on the same thread as the rebalance handler meaning we can deadlock if we get unlucky waiting on a poll. The solution here would be to build a different consumer around BaseConsumer that doesn't have this problem and re-implement the countdown latch (there's a commit in here somewhere with it). For Snuba, not draining was safe, as Snuba explicitly doesn't drain today and is OK with duplicate work.

rust-arroyo/src/processing/stream/stage.rs
tryangul: I haven't sketched it out, but I think the watermark collector is a good idea. Maybe each branch writes to an mpsc that goes to the collector. I'd want to hide the dirty part behind the API. I generally try to crib from the design of Reactor and other JVM reactive libraries when possible (Rust / Tokio aren't as mature sometim)

rust-arroyo/src/processing/stream/source.rs
tryangul: `OffsetTracker::try_commit()` calls this which is what bubbles up the error. In Rust fashion, there's a subtle chain of `?` which eventually gets to `PipelineRunner::run()` which exits, the program shuts down and it gets restarted by kube. 

In the Push `run_once` we `unwrap()` the result of the commit immediately which should have the same result.

rust-arroyo/src/processing/stream/source.rs
tryangul: 1. I think this is another compelling reason to write a new stream consumer (rdkafka's StreamConsumer hides all this from us) where we can measure this better.
2. We could add it to the stage metrics, but need to think about how to do that performantly.
3. I think the per stage counters effectively give us this. We increment them as data passes through each, so we can see when and where we are stuck

PR #561, opened 2026-08-25
feat(reduce): track batch trigger reason on batch_time metric.

tryangul:
## What
Tag `arroyo.strategies.reduce.batch_time.ms` with which trigger (`size` or `time`) caused the flush.

## How
- Compute `size_trigger_complete`/`time_trigger_complete` explicitly in `Reduce::flush()`
- Add a `trigger_reason` tag to the existing `batch_time.ms` timer instead of a new metric

## Why
- We're doing some batch size in Snuba tuning and would like to know which trigger is actually firing
- Reusing the existing timer gets flush count *and* duration split by reason for free, no new metric to track

## Notes
- Known gap: forced flushes (shutdown/rebalance via `join()`) that hit neither trigger still get tagged `"time"` — fine for now, but a `"forced"` value would make the metric fully trustworthy

## Links
- [EAP-694](https://linear.app/getsentry/issue/EAP-694/revisit-byte-based-max-batch-size-calculation)

### DIFF
--- a/arroyo/processing/strategies/buffer.py
+++ b/arroyo/processing/strategies/buffer.py
@@ -42,6 +42,14 @@ class BufferProtocol(Protocol[TPayload, TResult]):
         """
         ...
 
+    @property
+    def readiness_reason(self) -> str:
+        """Returns why is_ready returned True.
+
+        Only meaningful when is_ready is True. Used for observability.
+        """
+        ...
+
     def append(self, message: BaseValue[TPayload]) -> None:
         """Accept a TPayload mutating the internal state of the batch builder."""
         ...
@@ -115,9 +123,12 @@ class Buffer(
             )
         )
         self.__next_step.submit(buffer_msg)
+
+        flush_reason = self.__buffer.readiness_reason if self.__buffer.is_ready else "force"
         self.__metrics.timing(
             "arroyo.strategies.reduce.batch_time",
             time.time() - self.__init_time,
+            tags={"flush_reason": flush_reason},
         )
 
         # Reset to the empty state.
--- a/rust-arroyo/src/processing/strategies/reduce.rs
+++ b/rust-arroyo/src/processing/strategies/reduce.rs
@@ -191,14 +191,19 @@ impl<T, TResult> Reduce<T, TResult> {
         }
 
         let batch_time = self.batch_state.batch_start_time.elapsed();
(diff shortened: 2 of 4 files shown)


rust-arroyo/src/processing/strategies/reduce.rs
```diff
@@ -191,14 +191,19 @@ impl<T, TResult> Reduce<T, TResult> {
...
         if !batch_complete && !force {
             return Ok(());
         }
 
-        timer!("arroyo.strategies.reduce.batch_time.ms", batch_time);
+        timer!(
+            "arroyo.strategies.reduce.batch_time.ms",
+            batch_time,
+            "trigger_reason" => if size_trigger_complete { "size" } else { "time" }
+        );
```
fpacifici, line 213: Could you please do this in the python version as well? We try to keep the observability compatible.

fpacifici (changes requested):
Please see my comment below

untitaker (approved):

rust-arroyo/src/processing/strategies/reduce.rs
tryangul: Yup will do

tryangul: Ok added, but it's less encapsulated by nature of the python design. I'll tag you on the relevant parts.

arroyo/processing/strategies/buffer.py
```diff
@@ -42,6 +42,14 @@ def is_ready(self) -> bool:
         """
         ...
 
+    @property
+    def readiness_reason(self) -> str:
+        """Returns why is_ready returned True.
+
+        Only meaningful when is_ready is True. Used for observability.
+        """
+        ...
```
tryangul, line 51: Only the buffer knows why it's "ready", yet the metric is recorded so it must get propagated via an interface change. I went with an @property method, but changing the return type of `is_ready` is another option.

The third option is an entirely new Protocol to propagate this.

It's not clear if this is entirely internal or a breaking API change. Greping in Sentry, I didn't find any implementors (hard to check since protocols are structural interfaces). Please let me know if we can proceed with this approach and how I should version it.

tryangul: @fpacifici

fpacifici (approved):

PR #562, opened 2026-08-26
chore: Widens supported rdkafka range to >=0.37.0,<0.40

lvthanh03:
Widens the supported rdkafka version range to <0.40. Arroyo builds successfully and all Rust and Kafka integration tests pass locally with rdkafka 0.39.0.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -22,14 +22,20 @@ jobs:
       - run: cargo fmt --check
 
   test:
-    name: "Tests"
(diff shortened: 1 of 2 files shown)


untitaker (approved):
it would be nice to have a test matrix in CI for those ranges

lvthanh03:
> it would be nice to have a test matrix in CI for those ranges

Yeah that makes sense. I also went with explicit version numbers instead of a pre-CI job that uses regex to parses version range which I think is just too much complexity

untitaker (approved):

PR #573, opened 2026-09-14
Add ReloadingKafkaProducer

untitaker:
POC of a kafka producer that can change config at runtime. It is built entirely outside of KafkaProducer so the POC is easier to review can be safely rolled out, but I'm not sure that's a good idea (after all, relay will probably be tied to KafkaProducer).

The basic idea of hot-reloading is that whenever a new config is provided, the producer is swapped out for a new one and the old one is drained in a bg thread. Bursts of config updates are debounced because only one swapping+draining operation can happen at a time. Newer config updates overwrite old ones.

`produce()` waits for the old producer to finish draining _iff_ the produced message has a partition key. A partition key implies strict ordering of messages per-partition and in order to preserve these semantics, it's technically more correct to wait for the old producer to drain before using the new producer. In practice Relay sets partition keys for every message regardless of ordering requirements, so there's also a producer setting to opt out of blocking drain again.

Config is fully hidden in Arroyo, users are expected to pass opaque config blobs from options-automator. These config blobs are ultimately managed by the streaming team in a central location.

ref STREAM-1804


rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1530 @@
...
+        let mut in_flight = self.inner.drains_in_flight.lock();
+        *in_flight -= 1;
+        if *in_flight == 0 {
+            self.inner.drain_done.notify_all();
+        }
+    }
+}
+
+impl ArroyoProducer<KafkaPayload> for ReloadingKafkaProducer {
+    fn produce(
```
untitaker, line 676: joris: this method is now blocking, we need to make it non-blocking (but not async) again. wait for drain in bg too

rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1530 @@
...
+    /// Hands a new config blob to the producer.
+    ///
+    /// Call when the blob changes. Unchanged blobs do nothing; invalid blobs
+    /// leave the current client untouched.
+    ///
+    /// This parses and hands off without blocking. [`ReloadOutcome::Accepted`]
+    /// means rollout started, not that the config is live; watch
+    /// [`ReloadingKafkaProducer::generation`]. Broker work happens on the worker.
+    pub fn push_config(&self, blob: &[u8]) -> Result<ReloadOutcome, ReloadError> {
+        let config = ConfigBlob::parse(blob)?.producer_config(&self.inner.selector)?;
```
untitaker, line 284: we need to add metrics here for failing to parse configs, tagged too

rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1530 @@
...
+            }
+        }
+    }
+
+    /// Checks that a newly built client can actually reach a broker.
+    ///
+    /// Client construction does no I/O; this prevents swapping in an
+    /// unreachable producer.
+    fn probe(&self, candidate: &KafkaProducer) -> Result<(), KafkaError> {
+        let Some(timeout) = self.settings.probe_timeout else {
```
evanh, line 456: Should we skip the probe if the timeout isn't set? Or block indefinitely?

evanh (commented):
Conceptually all makes sense to me!

rust-arroyo/src/backends/kafka/reloading_producer.rs
untitaker: i hate both, i made it non-optional

rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1681 @@
...
+
+impl Drop for ShutdownOnDrop {
+    fn drop(&mut self) {
+        self.inner.reload.lock().shutdown = true;
+        self.inner.reload_wake.notify_all();
+    }
+}
+
+struct Inner {
+    /// Read on every produce, taken for writing only to start buffering and
```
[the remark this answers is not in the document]
untitaker: yes i considered handling this but thought it was too much work. we'll have to deal with it at query time anyway since processes can die by e.g. OOM too. in the end we will be able to see an absence of datapoints.

rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1681 @@
...
+    }
+
+    #[test]
+    fn test_identical_config_does_not_swap() {
+        let cluster = MockCluster::new(1).unwrap();
+        let servers = cluster.bootstrap_servers();
+        let producer = producer(&servers);
+
+        let outcome = producer.push_config(&blob(&servers, "all")).unwrap();
+
```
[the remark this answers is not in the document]
untitaker: there are more bugs like this, adding a seq counter only fixes one of them. i want to punt on that rabbithole for now

evanh (approved):
Seems good to me, minus the couple followup TODOs. Nice work!

rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1681 @@
...
+                metrics::counter!(
+                    "arroyo.producer.config_reload_replay_failed",
+                    "topic" => self.selector.topic().to_owned(),
+                    "producer_name" => self.producer_name.clone(),
+                )
+                .increment(1);
+                tracing::error!(
+                    topic = %self.selector.topic(),
+                    %error,
+                    "could not replay a buffered message onto the new client"
```
lvthanh03, line 747: Does that mean we lose the message during replay if produce returns an error? Should we retry here?

rust-arroyo/src/backends/kafka/reloading_producer.rs
```diff
@@ -0,0 +1,1681 @@
...
+        }
+
+        let purged = old.in_flight_count().max(0) as u64;
+
+        #[cfg(test)]
+        self.purged_at_drop.store(purged, Ordering::SeqCst);
+
+        if purged > 0 {
+            metrics::counter!(
+                "arroyo.producer.config_reload_purged_messages",
```
lvthanh03, line 606: Should we drop the messages outright when `drain` times out? Could we retain the old client and retry draining, though that would further delay the reload.

rust-arroyo/src/backends/kafka/reloading_producer.rs
untitaker: yeah unfortunately (note that this is only for semantic partitioning, if you set `ignore_key_ordering`, the buffer is disabled entirely). rdkafka retries internally already though.

rust-arroyo/src/backends/kafka/reloading_producer.rs
untitaker: same as in the other comment, rdkafka already ought to retry on network errors, so i think the draining retry won't do much

